PR #16, opened 2019-05-06
Horizon rename

hendrikbrandt:
Tested with rosbags. I don't consider a full test on a robot necessary.


bitbots_vision/src/bitbots_vision/dynamic_color_space.py
```diff
@@ -302,8 +302,8 @@ class Heuristic:
     def __init__(self, debug_printer):
         # type: (DebugPrinter) -> None
         """
-        Filters new color space colors according to their position relative to the horizon.
-        Only colors that occur under the horizon and have no occurrences over the horizon get picked.
+        Filters new color space colors according to their position relative to the field_boundary.
```
JanGut: In comments like this where the word "field_boundary" is not used as a specific term it should be written without the under_score.

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -412,105 +412,105 @@ def _precise_horizon(self):
...
+    def candidate_under_field_boundary(self, candidate, y_offset=0):
         # type: (tuple, int) -> bool
         """
-        returns whether the candidate is under the horizon or not
+        returns whether the candidate is under the field_boundary or not
         :param candidate: the candidate, a tuple (upleft_x, upleft_y, width, height)
-        :param y_offset: an offset in y-direction (higher offset allows points in a wider range over the horizon)
-        :return: whether the candidate is under the horizon or not
+        :param y_offset: an offset in y-direction (higher offset allows points in a wider range over the field_boundary)
+        :return: whether the candidate is under the field_boundary or not
```
JanGut: Another example of the problem above.

bitbots_vision/src/bitbots_vision/dynamic_color_space.py
NFiedler: I think that this is not a crucial problem. Let's change these two occurrences and correct others when we see them.

hendrikbrandt: 'field_boundary' has been changed to 'field boundary' and 'field_boundary-mask' to 'field-boundary-mask' in comments

PR #33, opened 2019-06-20
Add time stamp delay parameter

Flova:
To fix the timing issue with the Basler cam timings I integrated a time delay parameter in the white balancer.
The Parameter can be easily be changed via dynamic reconfigure.

### DIFF
--- a/white_balancer/cfg/white_balancer.cfg
+++ b/white_balancer/cfg/white_balancer.cfg
@@ -6,6 +6,6 @@ from dynamic_reconfigure.parameter_generator_catkin import ParameterGenerator, i
 gen = ParameterGenerator()
 
 gen.add("temp", int_t, 1, "Color temp for white balance.", min=1000, max=12000)
-gen.add("delay", double_t, 1, "Delay the image time stamp by that value.", min=-2.0, max=2.0)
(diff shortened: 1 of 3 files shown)


white_balancer/cfg/white_balancer.cfg
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env python
 # configuration
 
-from dynamic_reconfigure.parameter_generator_catkin import ParameterGenerator, int_t
+from dynamic_reconfigure.parameter_generator_catkin import ParameterGenerator, int_t, double_t
 
 gen = ParameterGenerator()
 
 gen.add("temp", int_t, 1, "Color temp for white balance.", min=1000, max=12000)
+gen.add("delay", double_t, 1, "Delay the image time stamp by that value.", min=-2.0, max=2.0)
```
NFiedler: Delay is the wrong word here. Maybe change the sign?

Flova: I renamed delay to offset.

Flova:
> While I don't like that the offset is subtracted instead of added, I think that it should be fine for now.

Done ;)

Issue #36, opened 2019-06-21
Reset line point count to default


JanGut:
To save some resources while the RoboCup world championship 2019, the parameter `line_detector_linepoints_range` has been set to `0`.
This parameter determines the number of created line points.

Please reset this afterward to default `200`.

JanGut:
The `line_detector_linepoints_range` value is still 10000. We should evaluate this with default settings in f031 and probably again after we use a localization.

Flova:
See pull request #147

Flova:
See pull request

PR #37, opened 2019-06-21
Turn off line detection

JanGut:
Turn off line detection by creating zero line points.
Currently line points are not needed for further processing, therefore some resources can be saved.

#36 is a reminder to reset this to default.


JanGut:
> Looks fine, but please test this before merging. I didn't check for any division by zero errors or stuff like that.

Now tested and working in docker.

PR #40, opened 2019-06-21
Add camera parameter to launch file

JanGut:
This adds a new `camera` parameter to launch file. This parameter determines wether an image provider will be launched (as the `sim` parameter before).

This changes also the behavior of the `debug` parameter: if true, this also activates the publishing of several image messages such as field masks.


JanGut:
Please test all parameters before merging.

JanGut:
All parameters are tested and working.

PR #45, opened 2019-06-28
Debug images

JanGut:
This PR removes the `vision_debug_image` parameter which activated a simple image to show the debug image. This functionality is no longer needed, since the debug image can be published via a Image message.

This PR also adds the possibility of publishing the ball fcnn debug image by activating the launch parameter `debug`.


bitbots_vision/cfg/Vision.cfg
```diff
@@ -40,8 +40,8 @@ group_white_color_detector = group_color_detector.add_group("white_color_detecto
 group_red_color_detector = group_color_detector.add_group("red_color_detector")
 group_blue_color_detector = group_color_detector.add_group("blue_color_detector")
 
+group_ball_fcnn.add("ball_fcnn_publish_debug_img", bool_t, 0, "toggles publishing of the fcnn heatmap image for debug purposes", None)
```
timonegk: This parameter is added twice (group_ball_fcnn and group_ROS). I would prefer to put it in group_ROS only since the debug image setting is also set there

bitbots_vision/src/bitbots_vision/vision_modules/fcnn_handler.py
```diff
@@ -19,15 +19,16 @@ class FcnnHandler(CandidateFinder):
...
+        publish_debug_img: false  # toggles publishing of the fcnn heatmap image for debug purposes
+        model_path: '/models/2019_05_timon_basler'
+        threshold: .6  # minimal value for a candidate to be considered
         expand_stepsize: 4
         pointcloud_stepsize: 10
-        shuffle_candidate_list: true  # shuffles the list of possible candidate points
-        min_candidate_diameter: 10
-        max_candidate_diameter: 200
-        candidate_refinement_iteration_count: 1
+        shuffle_candidate_list: false  # shuffles the list of possible candidate points
```
NFiedler: Why is this false now?

bitbots_vision/src/bitbots_vision/vision.py
```diff
@@ -321,10 +319,8 @@ def _dynamic_reconfigure_callback(self, config, level):
         self._ball_candidate_threshold = config['vision_ball_candidate_rating_threshold']
         self._ball_candidate_y_offset = config['vision_ball_candidate_field_boundary_y_offset']
 
-        self.debug_image = config['vision_debug_image']
-        self.debug_image_msg = config['vision_publish_debug_image']
-        self.debug = self.debug_image or self.debug_image_msg
-        if self.debug:
+        self.do_publish_debug_image = config['vision_publish_debug_image']
```
NFiedler: the "do_"-part seems weird...

bitbots_vision/src/bitbots_vision/vision.py
```diff
@@ -302,11 +302,9 @@ def handle_image(self, image_msg):
...
             #     self.line_detector.get_nonlinepoints(),
             #     (0, 0, 0))
-            # debug_image_dings.draw_line_segments(line_detector.get_linesegments(), (180, 105, 255))
-            if self.debug_image:
-                self.debug_image_dings.imshow()
-            if self.debug_image_msg:
-                self.pub_debug_image.publish(self.bridge.cv2_to_imgmsg(self.debug_image_dings.get_image(), 'bgr8'))
+
+            # publish debug image
+            self.pub_debug_image.publish(self.bridge.cv2_to_imgmsg(self.debug_image_dings.get_image(), 'bgr8'))
```
NFiedler: You still need an "if" here.

bitbots_vision/src/bitbots_vision/vision_modules/fcnn_handler.py
JanGut: Because you have set this to false a year ago and I just updated the example config to our default values...

bitbots_vision/src/bitbots_vision/vision.py
JanGut: No, because the `if`-clause in line 258  decides whether a debug image should be drawn and also published.

bitbots_vision/src/bitbots_vision/vision.py
JanGut: resolved with caca46c0eb4cdbe93fae8d45f04c48d7d2bcf00b.

PR #46, opened 2019-06-29
dynamic reconfigure: generate enums for models and color spaces

timonegk:
As of this Pull Request, enums for dynamic reconfigure are generated automatically to represent the contents of the respective folder. This has the advantage that typos are avoided and a better overview of the possible values is given. The only drawback is that the vision has to be rebuilt whenever a new color space or model is added.
Since a default has to be selected for enums in dynamic reconfigure, I chose the first file in the folder because it will always be overridden by the default value in the YAML configuration.
This Pull Request was done with the help of @JanGut

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,14 @@ README
 ======
 
 This is the vision package of the Hamburg Bit-Bots.
-For standardized camers such as USB-webcams we use the wolves_image_provider package as image provider.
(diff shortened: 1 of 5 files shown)


NFiedler:
Please test it on a robot before merging.

JanGut:
Tested on robot. Works as expected.

PR #78, opened 2019-08-19
Reorganization of color space files

JanGut:
This PR resolves issue #71.

Old interpolated YAML files have been deleted. Other old YAML file have been moved into a subfolder.
Interpolated pickle files of all (relevant) color spaces have been created.
The ending of all pickle color space files has changed from `.txt` to `.pickle`.

(Currently using pickles throws an error even on the master...)


bitbots_vision_tools/scripts/colorspace_tool.py
```diff
@@ -290,7 +290,7 @@ def save(self, filename):
             blue = blue
         )
 
-        filename = '{}_{}.txt'.format(filename, output_type)
+        filename = f'{filename}_{output_type}.pickle'
```
Flova: This syntax is cool, but only supported in python3

bitbots_vision/src/bitbots_vision/dynamic_color_space.py
```diff
@@ -229,7 +229,7 @@ def queue_to_color_space(self, queue):
         for new_color_value_list in queue:
             color_space = np.append(color_space, new_color_value_list[:,:], axis=0)
         # Return a color space, which contains all colors from the queue
-        return color_space
+        return color_space.astype(int)
```
Flova: Does this belong to this pull request?

Flova:
Now you interpolated all color spaces. In some scenarios, the interpolation is not wanted. Also, the smaller YAML colorspaces seem to remain. I thought we switch completely from the YAML to the pickle format. This would require the conversion of the remaining color spaces. The Colorspace tool is able to convert them without touching anything.

Flova:
When all color spaces are in the pickle format we could also drop the YAML color space support and reduce complexity.

bitbots_vision_tools/scripts/colorspace_tool.py
JanGut: Well, the shebang says py3, but i can change this

JanGut:
> The Colorspace tool is able to convert them without touching anything.

Yeah, in my case, the pickle files were all empty and contained no colors.

ftsell:
Allow me to give my senf too: You can move files with `git mv` instead of just `mv`. This should mark files explicitly as _moved_ and not _deleted and created_.
It's just something I noticed while scrolling over the changes

JanGut:
Thank you for the information, @ftsell. I will try it the next time!
I am closing this PR, since I have to remake all pickle files again.

Issue #90, opened 2019-09-13
Use thread-safety main worker in loop


jaagut:
To reduce errors due to dynamic reconfiguration, we should use a thread-safety main worker in the main loop.

Flova:
Fixed with #97 in the `vision.py` still TODO for the Dynamic Color Space node. Maybe post version 1.

JanGut:
We should also consider this instead of globals for the dynamic color detector...

Issue #96, opened 2019-09-16
Memory leak


Flova:
There is some weird behavior, which occurs when the vision gets reconfigured. With each callback the frame rate drops dramatically, no matter which parameter has been changed. In the debugging process I also discovered a small memory leak which occurs while reconfiguration in the described matter.

Flova:
This issue is not occurring anymore. Maybe it was something with the dependencies on the Jetson.

PR #97, opened 2019-09-16
Vision parallel

Flova:
When I discovered the Issue #96 I tried to fix it by improving the thread-safety. This resulted in this pull request. It also fixes issue #90 .

### DIFF
--- a/bitbots_vision/src/bitbots_vision/dynamic_color_space.py
+++ b/bitbots_vision/src/bitbots_vision/dynamic_color_space.py
@@ -138,8 +138,8 @@ class DynamicColorSpace:
 
         # Drops old images
         image_age = rospy.get_rostime() - image_msg.header.stamp
-        if 0.1 < image_age.to_sec() < 1000.0:
-            rospy.loginfo('Vision: Dropped incoming Image-message, because its too old!')
+        if 1.0 < image_age.to_sec() < 1000.0:
+            rospy.logwarn_throttle(2, 'Vision: Dropped incoming Image-message, because its too old! ({} sec)'.format(image_age.to_sec()))
             return
 
         self.handle_image(image_msg)
--- a/bitbots_vision/src/bitbots_vision/vision.py
+++ b/bitbots_vision/src/bitbots_vision/vision.py
@@ -98,7 +98,7 @@ class Vision:
 
     def main_loop(self):
         """
-        Main loop that processed the images and configuration changes
+        Main loop that processes the images and configuration changes
         """
         while not rospy.is_shutdown():
             # Lookup if there is another configuration available
--- a/bitbots_vision/src/bitbots_vision/vision.py
+++ b/bitbots_vision/src/bitbots_vision/vision.py
@@ -171,17 +171,22 @@ class Vision:
 
         # Check if the dynamic color space field color detector or the static field color detector should be used
         if self._use_dynamic_color_space:
-            # Set dynamic color space field color detector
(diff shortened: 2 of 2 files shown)


bitbots_vision/src/bitbots_vision/vision.py
```diff
@@ -86,17 +93,50 @@ def __init__(self):
...
-        rospy.spin()
+        # Run the vision main loop
+        self.main_loop()
+
+    def main_loop(self):
+        """
+        Main loop that processed the images and configuration changes
+        """
+        while not rospy.is_shutdown():
+            # Lookup if theres another configuration avalabile
```
JanGut:
```suggestion
            # Lookup if there is another configuration available
```

bitbots_vision/src/bitbots_vision/dynamic_color_space.py
```diff
@@ -138,8 +138,8 @@ def image_callback(self, image_msg):
 
         # Drops old images
         image_age = rospy.get_rostime() - image_msg.header.stamp
-        if 0.1 < image_age.to_sec() < 1000.0:
-            rospy.loginfo('Vision: Dropped incoming Image-message, because its too old!')
+        if 1.0 < image_age.to_sec() < 1000.0:
+            rospy.logwarn_throttle(2, 'Vision: Dropped incoming Image-message, because its too old! ({} sec)'.format(image_age.to_sec()))
```
NFiedler: Is this related to the PR?

bitbots_vision/src/bitbots_vision/vision.py
```diff
@@ -86,17 +93,50 @@ def __init__(self):
...
+                self.configure_vision(*reconfigure_data)
+            # Check if a new image is avalabile
+            elif self.transfer_image_msg is not None:
+                # Copy image from shard memory
+                image_msg = deepcopy(self.transfer_image_msg)
+                self.transfer_image_msg = None
+                # Run the vision pipeline
+                self.handle_image(image_msg)
+                # Now the first image has been processed
+                self._first_image_callback = False
```
NFiedler: do we need a sleep here i the else part?

bitbots_vision/src/bitbots_vision/dynamic_color_space.py
Flova: No not really, I was wondering why this showed up here. It is from another pull request yet merged and closed pull request.

bitbots_vision/src/bitbots_vision/vision.py
Flova: You mean, that the loop should not spin at insanely high rates if there's nothing to? We cloud add it, although I didn't noticed any weird behavior in the testing.

bitbots_vision/src/bitbots_vision/vision.py
```diff
@@ -86,17 +93,50 @@ def __init__(self):
...
+                # Copy config from shard memory
+                self.transfer_reconfigure_data_read_flag = True
+                reconfigure_data = deepcopy(self.transfer_reconfigure_data)
+                self.transfer_reconfigure_data_read_flag = False
+                self.transfer_reconfigure_data = None
+                # Run vision reconfiguration
+                self.configure_vision(*reconfigure_data)
+            # Check if a new image is avalabile
+            elif self.transfer_image_msg is not None:
+                # Copy image from shard memory
```
johagge:
```suggestion
                # Copy image from shared memory
```

PR #100, opened 2019-09-16
Switch to goal parts

Flova:
Switch to goal parts instead of goal msg. Do not merge before this is implemented in the behavior/transformer correctly.


Flova:
> And talk to @MonsterUntermBett abo.ut this so it doesn't break all her stuff.
I talked to her and it looks fine

Issue #61, opened 2019-07-20
Evaluate the ros log levels


Flova:
We should evaluate the ROS log levels. This is related to #18.

JanGut:
Resolved with #108.

PR #108, opened 2019-09-18
Change log levels

Flova:
Changes some log levels. Also evaluated all the other log levels. Related to issue #61.

### DIFF
--- a/bitbots_vision/src/bitbots_vision/vision.py
+++ b/bitbots_vision/src/bitbots_vision/vision.py
@@ -153,7 +153,7 @@ class Vision:
         # Print, if the vision uses the sim color or not
         if ros_utils.config_param_change(self.config, config, 'vision_use_sim_color'):
             if config['vision_use_sim_color']:
-                rospy.loginfo('Loaded color space for SIMULATOR.')
+                rospy.logwarn('Loaded color space for SIMULATOR.')
             else:
                 rospy.loginfo('Loaded color space for REAL WORLD.')
 
--- a/bitbots_vision/src/bitbots_vision/vision.py
+++ b/bitbots_vision/src/bitbots_vision/vision.py
@@ -114,7 +114,7 @@ class Vision:
         if ros_utils.config_param_change(self.config, config, 'vision_publish_debug_image'):
             self._publish_debug_image = config['vision_publish_debug_image']
             if self._publish_debug_image:
-                rospy.logwarn('Debug images are enabled')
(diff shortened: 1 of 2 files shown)


bitbots_vision/src/bitbots_vision/vision.py
```diff
@@ -114,46 +114,46 @@ def _dynamic_reconfigure_callback(self, config, level):
...
+                rospy.loginfo('HSV mask image publishing is enabled')
             else:
                 rospy.loginfo('HSV mask image publishing is disabled')
 
         # Should the (dynamic color space-) field mask image be published?
         if ros_utils.config_param_change(self.config, config, 'vision_publish_field_mask_image'):
             self._publish_field_mask_image = config['vision_publish_field_mask_image']
             if self._publish_field_mask_image:
-                rospy.logwarn('(Dynamic color space-) Field mask image publishing is enabled')
+                rospy.loginfo('(Dynamic color space-) Field mask image publishing is enabled')
```
NFiedler: I'd be okay with all of the above being warnings as the image publishing is an overhead we really do not want in a game.

bitbots_vision/src/bitbots_vision/vision.py
```diff
@@ -114,46 +114,46 @@ def _dynamic_reconfigure_callback(self, config, level):
...
             else:
                 rospy.loginfo('(Dynamic color space-) Field mask image publishing is disabled')
 
         # Print, if the vision uses the sim color or not
         if ros_utils.config_param_change(self.config, config, 'vision_use_sim_color'):
             if config['vision_use_sim_color']:
-                rospy.logwarn('Loaded color space for SIMULATOR.')
+                rospy.loginfo('Loaded color space for SIMULATOR.')
```
NFiedler: this should be a warning as it could result in really weird behavior if you expect that the other color space is loaded.

bitbots_vision/src/bitbots_vision/vision.py
timonegk: There is no overhead because ROS will not publish anything when nobody subscribes to the topic.

bitbots_vision/src/bitbots_vision/vision.py
Flova: Okay this makes sense

JanGut: Resolved with 49b4037c5230cea0d7d1cf42c5de71ab7a3cef38.

PR #118, opened 2019-09-25
Add auto build for vision extensions

Flova:
Now the vision extensions are included in the CMake build process.
Closes issue #77


JanGut:
I don't feel confident reviewing this...

Flova:
Maybe you can test it on your device.
E.g. see if it throws an error if you remove something in the extensions and if it succeeds if your revert the change.
Or add an print in the extensions and rebuild.

Flova:
But it's also fine if you think this does not belong to you.

PR #127, opened 2019-09-28
New doc system

Flova:
This pull request implements the new documentation system in the vision. 
It also fixes some wrong formatted docstrings and updates the vision package description.


JanGut:
I really like the new description! I'll make myself some additional thoughts about the content of the description.
Besides, we should write "ROS", "YOLO" and "FCNN" in caps.

I think we have a lot more of docstrings  to rewrite and correct...

bitbots_vision/src/bitbots_vision/vision_modules/candidate.py
```diff
@@ -12,135 +12,106 @@ def __init__(self, x1=0, y1=0, width=0, height=0, rating=None):
     def get_width(self):
         # type: () -> int
         """
-        returns...
```
JanGut: Even tho this is self-explanatory, I would also include some description like `Returns the width in <units> of the candidate.`.

Also in the cases below...

Flova: Putting it in the Description and the `:returns:` part is a duplication. But we can indeed improve the `returns` part.

PR #144, opened 2019-11-01
Caching parameter

Flova:
Adds parameter to deactivate the caching for profiling reasons.


bitbots_vision/src/bitbots_vision/vision_modules/ros_utils.py
```diff
@@ -465,14 +467,23 @@ def speak(string, speech_publisher):
     speak_message.text = string
     speech_publisher.publish(speak_message)
 
-def config_param_change(old_config, new_config, params):
+def set_global_params(params):
```
JanGut: I wouldn't call this `global`, because this could cause irritation regarding global variables.

bitbots_vision/cfg/Vision.cfg
```diff
@@ -174,5 +174,6 @@ group_debug.add("vision_publish_debug_image", bool_t, 0, "Publish debug image me
 group_debug.add("ball_fcnn_publish_debug_img", bool_t, 0, "Publish the fcnn heatmap image for debug purposes", None)
 group_debug.add("vision_publish_HSV_mask_image", bool_t, 0, "Publish all three HSV color detector mask image mesages for debug purposes ", None)
 group_debug.add("vision_publish_field_mask_image", bool_t, 0, "Publish field mask image message for debug purposes", None)
+group_debug.add("caching", bool_t, 0, "DO NOT TOUCH! Used to deactivate caching for profiling reasons", None)
```
JanGut: `Do not touch` is irritating. `Used to deactivate caching for profiling reasons` is enough. Why not allow deactivation of caching in the running system (for some reasons)?

bitbots_vision/src/bitbots_vision/vision_modules/ros_utils.py
Flova: Any better name, for a parameter, which is used by all detectors?

NFiedler: general parameters?

Flova: That sound nice 👍

bitbots_vision/cfg/Vision.cfg
timonegk: Why even add this setting to dynamic reconfigure?

Flova: Because it fits in the vision configuration pattern.

PR #146, opened 2019-11-02
Add new field boundary algorithm

Flova:
This adds a field boundary algorithm based on OpenCV resizing and blurring. 
It minimizes the python operations while accepting some OpenCV operation overhead. The approach is about 5 times faster on my PC vs. the normal reversed field boundary algorithm. The performance on the Jetson has to be evaluated tho.


bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -629,13 +657,59 @@ def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps,
...
+                if subsampled_mask[max_y, x_position] < int(green_threshold / 1000 * 255):
+                    # Reset to last step
+                    max_y += 1
+                    break
+
+            # Scale the stuff for the original image + black magic
+            field_boundary_points.append(
+                (int((x_position + 0.5) * (field_mask.shape[1] / x_steps)),
+                 int(max_y * (field_mask.shape[0] / y_steps))))
+        # Fix tuff at the image edges
```
NFiedler: "tuff".

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -629,13 +657,59 @@ def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps,
...
+        # Define the blur kernel
+        kernel = (2 * roi_height + 1, 2 * roi_width + 1)
+        # Blur the downscaled image to fill holes in the field mask
+        subsampled_mask = cv2.GaussianBlur(subsampled_mask, kernel, 0)
+
+        field_boundary_points = []
+
+        # Iterate horizontal over the image
+        for x_position in range(subsampled_mask.shape[1]):
+            # Iterate vertical over the downscaled mask
```
NFiedler: ...horizontally... ...vertically...

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -629,13 +657,59 @@ def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps,
...
+                    max_y += 1
+                    break
+
+            # Scale the stuff for the original image + black magic
+            field_boundary_points.append(
+                (int((x_position + 0.5) * (field_mask.shape[1] / x_steps)),
+                 int(max_y * (field_mask.shape[0] / y_steps))))
+        # Fix tuff at the image edges
+        field_boundary_points[0] = (0, field_boundary_points[0][1])
+        field_boundary_points[-1] = (field_mask.shape[1]-1, field_boundary_points[0][1])
```
NFiedler: Are you sure that you want `field_boundary_points[0][1] ` here instead of `field_boundary_points[-1][1]` ?

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -629,13 +657,59 @@ def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps,
...
+        :returns: list of field boundary points
+        """
+        # calculate the field_mask which contains 0 for non-green pixels and 255 for green pixels in the image
+        # index counting up from top to bottom and left to right
+        field_mask = field_color_detector.get_mask_image()
+
+        # Scale the image down
+        subsampled_mask = cv2.resize(field_mask,(x_steps,y_steps), interpolation=cv2.INTER_AREA)
+        # Define the blur kernel
+        kernel = (2 * roi_height + 1, 2 * roi_width + 1)
```
NFiedler: Why is this so large?

bitbots_vision/cfg/Vision.cfg
```diff
@@ -27,6 +27,7 @@ field_boundary_detector_enum = gen.enum(
     gen.const("iteration",str_t, "iteration", "searches from the top until green"),
     gen.const("dynamic",str_t, "dynamic", "uses head_joint_state messages to switch between methods"),
     gen.const("reversed",str_t, "reversed", "searches from the bottom until white"),
+    gen.const("cvreversed",str_t, "cvreversed", "searches from the bottom until white unsing opencv"),
```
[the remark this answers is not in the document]
Flova: A bit long, but nice idea. Let's go with that.

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
Flova: The kernel? This why should it be large? The roi_height is normally a value of 1. The rest of the calculation is added, to get an odd result every time, which is needed for the blur. Also this calculation output is 1, which means no bluring for an roi size of 0 and 1 on each side for roi 1 and so on.

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
Flova: Oh yeah I fix it

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
NFiedler: First, "one on each side for roi of 1" would mean that a "height" or "width" of 1 leads to an actual height or width of 3, which is larger than 1. But I think that this is the case everywhere. Second, in this case the width of the kernel considers the subsampled pixels which significantly enlarges the corresponding area in the original image.

Flova: Of cause the area is quite large after the downscaling. But the kernel normally only considers the direct neighbors by setting the value to 1. This works quite well on our settings and is much faster than bluring the original image. Higher values are nessesary if we increase the resolution further and need to blur bigger areas. Of cause height is a bit misleading if we a talk about some radius like thing, but an other parameter would break the interface quite a bit and this one was the closest one. 
If we want to use the real height/width, we could add an integer division on the height/width value first. That would also output only odd values for the kernel size, but parameter values like 2 and 3 would be equal.

bitbots_vision/cfg/Vision.cfg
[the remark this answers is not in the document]
Flova: Done

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -629,13 +657,59 @@ def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps,
         return _field_boundary_points
 
 
+class DownsamplingReversedFieldBoundaryAlgorithm(FieldBoundaryAlgorithm):
+    @staticmethod
+    def _calculate_field_boundary(image, field_color_detector, x_steps, y_steps, roi_height, roi_width, roi_increase, green_threshold):
+        # type: () -> list
```
JanGut: This is defiantly wrong.

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -629,13 +657,59 @@ def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps,
...
+        """
+        Finds the points of the field boundary visible in the image. Uses the reversed method iterating from bottom to
+        top until it finds enough non green points. Useful for when two fields are adjacent to each other.
+
+        :returns: list of field boundary points
+        """
+        # calculate the field_mask which contains 0 for non-green pixels and 255 for green pixels in the image
+        # index counting up from top to bottom and left to right
```
JanGut: Why is this here?

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -629,13 +657,59 @@ def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps,
...
+            for y_position in range(subsampled_mask.shape[0]):
+                # Invert the current vertical value
+                max_y = (subsampled_mask.shape[0] - 1) - y_position
+                # Check if we found a pixel under the threshold
+                if subsampled_mask[max_y, x_position] < int(green_threshold / 1000 * 255):
+                    # Reset to last step
+                    max_y += 1
+                    break
+
+            # Scale the stuff
```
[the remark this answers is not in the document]
Flova: I totally agree with you

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
Flova: I added the integer division thing.

bitbots_vision/cfg/Vision.cfg
```diff
@@ -27,6 +27,7 @@ field_boundary_detector_enum = gen.enum(
     gen.const("iteration",str_t, "iteration", "searches from the top until green"),
     gen.const("dynamic",str_t, "dynamic", "uses head_joint_state messages to switch between methods"),
     gen.const("reversed",str_t, "reversed", "searches from the bottom until white"),
+    gen.const("downsampling_reversed",str_t, "downsampling_reversed", "searches from the bottom until white unsing opencv"),
```
NFiedler: I don't like 'using opencv'

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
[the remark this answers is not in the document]
NFiedler: Are you serious???

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -629,13 +657,59 @@ def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps,
...
+            field_boundary_points.append(
+                (int((x_position + 0.5) * (field_mask.shape[1] / x_steps)),
+                 int(max_y * (field_mask.shape[0] / y_steps))))
+        # Fix stuff at the image edges
```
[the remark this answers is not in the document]
NFiedler: It still does.

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -557,16 +583,13 @@ def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps,
 class ReversedFieldBoundaryAlgorithm(FieldBoundaryAlgorithm):
     @staticmethod
     def _calculate_field_boundary(_image, _field_color_detector, _x_steps, _y_steps, _roi_height, _roi_width, _roi_increase, _green_threshold):
-        # type: () -> list
```
NFiedler: instead of removing these, you/someone should correct them. In a separate PR.

bitbots_vision/cfg/Vision.cfg
Flova: Should be better now

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
Flova: Yea, but now they are definitely incorrect and only on very few functions in the whole vision. We should add them for each function.

PR #148, opened 2019-11-02
HSV field color detector

### LINKED ISSUE #82: HSV green detector for demo purposes
As discussed in #59 the parameter gets removed but should be added with a propper implementation later on.

Flova:
This pull request adds an HSV field color detector, for faster configurations e.g. at demonstrations. Closes #82.

### DIFF
--- a/bitbots_vision/cfg/Vision.cfg
+++ b/bitbots_vision/cfg/Vision.cfg
@@ -121,12 +121,12 @@ group_neural_network_type.add("ball_candidate_rating_threshold", double_t, 0, "A
 group_field_color_detector.add("field_color_detector_path", str_t, 0, "Color space for the field color detector", dummy_list[0], edit_method=field_color_space_enum)
 
 group_hsv_field_color_detector.add("field_color_detector_use_hsv", bool_t, 0, "Should the dummy green HSV detector be used instead", None)
-group_hsv_field_color_detector.add("green_color_detector_lower_values_h", int_t, 0, "Lower bound for the green color detector hue", min=0, max=180)
-group_hsv_field_color_detector.add("green_color_detector_lower_values_s", int_t, 0, "Lower bound for the green color detector saturation", min=0, max=255)
-group_hsv_field_color_detector.add("green_color_detector_lower_values_v", int_t, 0, "Lower bound for the green color detector value/brightness", min=0, max=255)
-group_hsv_field_color_detector.add("green_color_detector_upper_values_h", int_t, 0, "Upper bound for the green color detector hue", min=0, max=180)
-group_hsv_field_color_detector.add("green_color_detector_upper_values_s", int_t, 0, "Upper bound for the green color detector saturation", min=0, max=255)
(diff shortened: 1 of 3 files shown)


bitbots_vision/src/bitbots_vision/vision.py
```diff
@@ -232,6 +232,17 @@ def _configure_vision(self, config, level):
                     config,
                     self._package_path)
 
+        # Check if params changed
+        if ros_utils.config_param_change(self._config, config,
+                r'^green_color_detector_|field_color_detector_use_hsv'):
+            # Deactivate dynamic color space
```
JanGut: As this is a duplicate of the code above, I suggest outsourcing this.

bitbots_vision/cfg/Vision.cfg
```diff
@@ -119,6 +120,14 @@ group_neural_network_type.add("ball_candidate_rating_threshold", double_t, 0, "A
...
 
+group_hsv_field_color_detector.add("field_color_detector_use_hsv", bool_t, 0, "Should the dummy green HSV detector be used instead", None)
+group_hsv_field_color_detector.add("green_color_detector_lower_values_h", int_t, 0, "Lower bound for the green color detector hue", min=0, max=180)
+group_hsv_field_color_detector.add("green_color_detector_lower_values_s", int_t, 0, "Lower bound for the green color detector saturation", min=0, max=255)
+group_hsv_field_color_detector.add("green_color_detector_lower_values_v", int_t, 0, "Lower bound for the green color detector value/brightness", min=0, max=255)
+group_hsv_field_color_detector.add("green_color_detector_upper_values_h", int_t, 0, "Upper bound for the green color detector hue", min=0, max=180)
+group_hsv_field_color_detector.add("green_color_detector_upper_values_s", int_t, 0, "Upper bound for the green color detector saturation", min=0, max=255)
+group_hsv_field_color_detector.add("green_color_detector_upper_values_v", int_t, 0, "Upper bound for the green color detector value/brightness", min=0, max=255)
```
NFiedler: why do you switch between field and green color detector?

bitbots_vision/src/bitbots_vision/vision.py
Flova: Which part is duplicated? The parameters in the are different in the both cases.

bitbots_vision/cfg/Vision.cfg
Flova: I thought, we have the field color detector, which is normally the pixellist one. Than we have the ability to use a hsv one instead. This hsv field color detector is basically a hsv green color detector, that could be used for other stuff. See the line detector, which uses a white color detector.

NFiedler: I think it is confusing and a bit misleading as we try to approximate the green color of the field specifically. Additionally, you add green color parameters to the field color group.

Flova:
We can rename it if you want.

PR #176, opened 2019-12-11
Add python colorpicker

### LINKED ISSUE #87: Rewrite color picker in Python
Creation of color space files is essential for the usage of our vision node. Currently, no documentation for those files (see #42 and bit-bots/bitbots_meta#31) and no tool for the creation of these is publicly available.

Therefore we should write a new color picker in Python maybe with integration into a running vision node, so that changes are visible immediately.
This tool should also include a undo function as well as the creation of negative color spaces. This tool should save color spaces as pickle (or legacy yaml) files.

Flova:
Adds the new colorpicker to the vision tools package.

## Proposed changes
Adds the `colorpicker.py` file.

## Related issues
Closes #87 

## Necessary checks
- [ ] Update package version
- [x] Run linters
- [x] Run `catkin build`
- [x] Write documentation
- [x] Test on your machine
- [ ] ~~Test on the robot~~


bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,230 @@
...
+
+def click_and_crop(event, x, y, flags, param):
+	"""
+	Callback for the OpenCV cursor.
+
+	:param event: Event type
+	:param x: Mouse x position
+	:param y: Mouse y position
+	:param flags: Callback flags
+	:param param: Some unused parameter
```
johagge: is this parameter necessary if is unused? 
If it is necessary would it make sense to initialize it with param=none, so calling the method is less confusing than having to call it with an unused parameter?

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def init_text():
+	"""
+	Prints some usage information.
+	"""
+	rospy.loginfo( "Welcome to the Colorpicker\n\
+Click to select the colors in the OpenCV window.\n\
+Use <shift> and <left click> to remove colors.\n\
+To undo a step press 'u'.\n\
```
johagge:
```suggestion
To undo a step press 'u'.\n\
To increase the size of the selection box press '+', to decrease press '-' \n\
```

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,230 @@
...
+		# Set to default
+		input_path = "~/.ros/color_space.pickle"
+
+	# Create abs path
+	input_path = os.path.abspath(os.path.expanduser(input_path))
+
+	# Get the color values
+	color_indices = np.where(history[-1] == 1)
+
+	# Create data struckture for the file
```
johagge:
```suggestion
	# Create data structure for the file
```

bitbots_vision_tools/scripts/colorpicker.py
Flova: It is indeed intended to use the new line characters here. This improves the readability in the terminal.

bitbots_vision_tools/scripts/colorpicker.py
Flova: This is a callback, so the parameters are required by OpenCV.

Flova: The param itself if not documented in the OpenCV doku.

bitbots_vision_tools/scripts/colorpicker.py
johagge: My suggested change was not supposed to change anything about the new lines. It was only supposed to add the second line which explains the '+' and '-' buttons.

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+        Callback for the OpenCV cursor.
+
+        :param event: Event type
+        :param x: Mouse x position
+        :param y: Mouse y position
+        :param flags: Callback flags
+        :param param: Some unused parameter
+        """
+        if flags == (cv2.EVENT_FLAG_SHIFTKEY + cv2.EVENT_FLAG_LBUTTON):
+            # Substracts colors
```
johagge:
```suggestion
            # Subtracts colors
```

bitbots_vision_tools/scripts/colorpicker.py
Flova: Ah okay.

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+
+            # Key checks for UI events
+            key = cv2.waitKey(1) & 0xFF
+
+            # Increase selection size
+            if key == ord("+"):
+                self._box_size += int(self._box_size * 0.2) + 1
+            # Reduce selection size
+            if key == ord("-") and self._box_size > 1:
+                self._box_size -= int(self._box_size * 0.2) - 1
```
JanGut: I assume, you also want:
```suggestion
                self._box_size -= int(self._box_size * 0.2) + 1
```

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+            # Undo
+            if key == ord("u") and len(self._history) > 1:
+                del self._history[-1]
+            # Quit exit
+            elif key%256 == 27 or 0xFF == ord('q'):
```
JanGut: Please explain this hex stuff or provide a reference...

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+
+        :param msg: Image message
+        """
+        self._image = self._cv_bridge.imgmsg_to_cv2(msg, 'bgr8')
+
+    def ui_loop(self):
+        """
+        Runs the UI window.
+
+        :return: Does the user exit the loop with a save flag set?
```
JanGut:
```suggestion
        :return bool: Flag for saving after exit.
```

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+        # Close all open windows
+        cv2.destroyAllWindows()
+        return save
+
+    def _save(self):
+        """
+        Saves the current colorspace in a pickle file.
+        """
+        # Get user input
+        input_path = input("Enter color space path: ")
```
JanGut:
```suggestion
        input_path = input("Enter color space file path: ")
```

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # Check if the user entered something
+        if input_path == "":
+            # Set to default
+            input_path = "~/.ros/color_space.pickle"
```
JanGut: I suggest to set this default path as a global in the top.

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+
```
JanGut:
```suggestion
                print("Saving to default path '{}'...".format(DEFAULT_SAVE_PATH))

```

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+        # Create abs path
```
JanGut:
```suggestion
        # Create absolute path
```

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # Save
+        with open(input_path, 'wb') as outfile:
+            pickle.dump(data, outfile, protocol=2)
+
+        rospy.loginfo("Output saved to '{}'.".format(input_path), logger_name="colorpicker")
+
+    def run(self):
+        """
+        This function runs the program
```
JanGut:
```suggestion
        Run the Colorpicker.
```

Flova:
We should add a max resolution for input images and eventually downscale large images...

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+
+        :param event: Event type
+        :param x: Mouse x position
+        :param y: Mouse y position
+        :param flags: Callback flags
+        :param param: Some unused parameter
+        """
+        if flags == (cv2.EVENT_FLAG_SHIFTKEY + cv2.EVENT_FLAG_LBUTTON):
+            # Subtracts colors
+            self._undo_click = True
```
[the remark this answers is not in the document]
Flova: > Please reset the click bools before or do something like:
>
> ``` self._undo_click = (flags == cv.EVENT_FLAG_SHIFTKEY + cv2.EVENT_FLAG_LBUTTON)```
> ```self._left_click = (not self._undo_click and event == cv2.EVENT_LBUTTONUP)```

Good point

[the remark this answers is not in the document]
Flova: But the reset in the main loop is still required, because of the different update rates. Otherwise, the main loop adds the color space multiple times until the next mouse callback arrives.

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+                        selected_matrix[:,1],
+                        selected_matrix[:,2]] = 1
+                else:
+                    print("This should never happen")
+
+                # Append new color space to self._history
+                self._history.append(color_space)
+
+                # Reset events
+                self._left_click, self._undo_click = False, False
```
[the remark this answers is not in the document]
Flova: See comment above.

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+            if key == ord("u") and len(self._history) > 1:
+                del self._history[-1]
+            # Quit exit
+            elif key%256 == 27 or 0xFF == ord('q'):
+                save = False
+                break
+            # Save exit
+            elif key == ord("s") or key == ord("w"):
+                save = True
+                break
```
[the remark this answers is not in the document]
Flova: No, as I said in person, this is the purpose of the main loop.

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+                (0, 255, 0), 2)
+
+            # Show canvas
+            cv2.imshow("Colorpicker", canvas)
+
+
+            # Key checks for UI events
+            key = cv2.waitKey(1) & 0xFF
+
+            # Increase selection size
```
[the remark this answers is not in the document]
Flova: Selection box?

bitbots_vision_tools/scripts/colorpicker.py
Flova: This is OpenCV UI keycode stuff, read the OpenCV docs if you want to know more about it, I think this does not belong here.

bitbots_vision_tools/scripts/colorpicker.py
Flova: This printout happens due to the output file location printout at the bottom.

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,255 @@
...
+
+class Colorpicker(object):
+    """
+    The bitbots_colorpicker node. Which is used to select colors that occur in the field.
+    """
+    def __init__(self):
+        rospy.init_node('bitbots_colorpicker')
+        rospy.loginfo('Initializing colorpicker...', logger_name="colorpicker")
+        self._cv_bridge = CvBridge()
+        self._image = np.zeros((300,300,3), dtype=np.uint8)
```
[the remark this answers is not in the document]
Flova: Done

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,254 @@
...
+                        selected_matrix[:,2]] = 0
+                # Check if we add stuff
+                elif self._left_click:
+                    # Add color values to color space
+                    color_space[
+                        selected_matrix[:,0],
+                        selected_matrix[:,1],
+                        selected_matrix[:,2]] = 1
+                else:
+                    print("This should never happen")
```
JanGut: I think, we should either describe more precisely _what should never happen?_ or delete this else branch, since it is impossible to enter this branch in the current state. And, if someone changes code inside the `Check for click event` block, they could add this else branch again, either to debug or add behavior...

I suggest, when we are confident enough to state, that this branch could not be entered, then we should trust ourself and remove this else block.

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,254 @@
...
+
+                # Append new color space to self._history
+                self._history.append(color_space)
+
+                # Reset events
+                self._left_click, self._undo_click = False, False
```
JanGut:
```suggestion
                self._left_click = self._undo_click = False
```

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,254 @@
...
+        # Close all open windows
+        cv2.destroyAllWindows()
+        return save
+
+    def _save(self):
+        """
+        Saves the current colorspace in a pickle file.
+        """
+        # Get user input
+        input_path = input( "Press enter to save look up table to '{}' "
```
JanGut: On every other occasion, its called color space, so...
```suggestion
        input_path = input( "Press enter to save color space file to '{}' "
```

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,254 @@
...
+        Prints some usage information.
+        """
+        rospy.loginfo(
+            "Welcome to the Colorpicker.\n"
+            "Click to select the colors in the OpenCV window.\n"
+            "Use <shift> and <left click> to remove colors.\n"
+            "To undo a step press 'u'.\n"
+            "To increase the size of the selection box press '+', to decrease press '-' \n"
+            "To save the color space press 's' followed by the path and <enter>.\n"
+            "Exit using <esc>.\n\n\n", logger_name="colorpicker")
```
JanGut: Maybe remove these many `\n` and instead add some spacers like `###########` before and after the _content_ lines.

bitbots_vision_tools/scripts/colorpicker.py
```diff
@@ -0,0 +1,254 @@
...
+        :param image: The image canvas
+        :param mask: The binary mask
+        :param color: Color bgr tuple
+        :param opacity: Opacity of the mask
+        """
+        # Make a colored image
+        colored_image = np.zeros_like(image)
+        colored_image[:, :] = tuple(np.multiply(color, opacity).astype(np.uint8))
+
+        # Compose debug image
```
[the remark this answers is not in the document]
Flova: Maybe we should name it a preview.

bitbots_vision_tools/scripts/colorpicker.py
Flova: These are independent states and this syntax is quite confusing.

bitbots_vision_tools/scripts/colorpicker.py
Flova: But technically it's a look up table. But we use color space all the time, so it doesn't matter. We can change it.

bitbots_vision_tools/scripts/colorpicker.py
Flova: I think the blank lines clean the console a bit.

PR #178, opened 2019-12-12
bitbots_vision: lines: Adds line detector offset for masks

JanGut:
Adds line detector offset for masks and now also supports negative offset.

## Proposed changes
The offset parameter was not yet implemented for line masks.
This is also needed for the RoboCup project.

## Related issues
None

## Necessary checks
- [ ] Update package version
- [X] Run linters
- [X] Run `catkin build`
- [ ] ~~Write documentation~~
- [X] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
+++ b/bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
@@ -68,14 +68,38 @@ class FieldBoundaryDetector(object):
         self._convex_field_boundary_points = None
         self._mask = None
 
-    def get_mask(self):
+    def get_mask(self, offset=0):
         # type: () -> np.array
         """
+        :param offest: A vertical field boundary offset shift
         :return: np.array
         """
         # Compute mask (cached)
         self._compute_mask()
-        return self._mask
+        return self._shift_field_boundary_mask(self._mask, offset)
+
+    def _shift_field_boundary_mask(self, mask, offset):
+        shape = mask.shape
+
+        if offset == 0:
+            return mask
+
+        elif offset < 0:
+            # Shift mask downwards
+            offset = min(-offset, shape[0]-1)
+            frame = np.zeros(shape, dtype=np.uint8)
+            # Add mask with offset
+            frame[offset:shape[0]-1] = mask[0:shape[0]-1-offset]
+
+        elif offset > 0:
+            # Shift mask upwards
+            offset = min(offset, shape[0]-1)
+            frame = np.ones(shape, dtype=np.uint8)
+            frame = frame * 255
+            # Add mask with offset
+            frame[0:shape[0]-1-offset] = mask[offset:shape[0]-1]
+
+        return frame
 
     def _compute_mask(self):
         # type: () -> None
--- a/bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
+++ b/bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
@@ -71,7 +71,7 @@ class FieldBoundaryDetector(object):
     def get_mask(self, offset=0):
         # type: () -> np.array
         """
-        :param offest: A vertical field boundary offset shift
+        :param offset: A vertical field boundary offset shift
         :return: np.array
         """
         # Compute mask (cached)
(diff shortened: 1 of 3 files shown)


bitbots_vision/src/bitbots_vision/vision_modules/lines.py
```diff
@@ -146,12 +146,34 @@ def get_line_mask(self):
...
+            frame[offset:shape[0]-1] = mask[0:shape[0]-1-offset]
+        elif offset > 0:
+            # Shift mask upwards
+            offset = min(offset, shape[0]-1)
+            frame = np.ones(shape, dtype=np.uint8)
+            frame = frame * 255
+            # Add mask with offset
+            frame[0:shape[0]-1-offset] = mask[offset:shape[0]-1]
+
+        return frame
```
Flova: This function should be part of the `get_mask()` function of the field boundary detector, so other detectors are able to use this shifted mask too. Maybe with the offset as an optional parameter.

bitbots_vision/src/bitbots_vision/vision_modules/lines.py
```diff
@@ -146,12 +146,34 @@ def get_line_mask(self):
...
 
+    def _shift_field_boundary_mask(self, mask, offset):
+        shape = mask.shape
+
+        if offset == 0:
+            return mask
+        elif offset < 0:
+            # Shift mask downwards
+            offset = min(-offset, shape[0]-1)
+            frame = np.zeros(shape, dtype=np.uint8)
```
Flova: Theoretically, we could shift the matrix in place and fill the missing stripe with zeros. This should increase the performance quite a lot.

bitbots_vision/src/bitbots_vision/vision_modules/lines.py
```diff
@@ -146,12 +146,34 @@ def get_line_mask(self):
...
+            # Add mask with offset
+            frame[offset:shape[0]-1] = mask[0:shape[0]-1-offset]
+        elif offset > 0:
+            # Shift mask upwards
+            offset = min(offset, shape[0]-1)
+            frame = np.ones(shape, dtype=np.uint8)
```
Flova: Same here but with ones.

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -68,14 +68,38 @@ def set_image(self, image):
...
+    def _shift_field_boundary_mask(self, mask, offset):
+        shape = mask.shape
+
+        if offset == 0:
+            return mask
+
+        elif offset < 0:
+            # Shift mask downwards
+            offset = min(-offset, shape[0]-1)
+            frame = np.zeros(shape, dtype=np.uint8)
```
Flova: Theoretically, we could shift the matrix in place and fill the missing stripe with zeros. This should increase the performance quite a lot.
(Moved from other file)

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -68,14 +68,38 @@ def set_image(self, image):
...
+            # Add mask with offset
+            frame[offset:shape[0]-1] = mask[0:shape[0]-1-offset]
+
+        elif offset > 0:
+            # Shift mask upwards
+            offset = min(offset, shape[0]-1)
+            frame = np.ones(shape, dtype=np.uint8)
```
Flova: Same here

bitbots_vision/src/bitbots_vision/vision_modules/lines.py
JanGut: This needs to be evaluated...

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
```diff
@@ -68,14 +68,38 @@ def set_image(self, image):
         self._convex_field_boundary_points = None
         self._mask = None
 
-    def get_mask(self):
+    def get_mask(self, offset=0):
         # type: () -> np.array
         """
+        :param offest: A vertical field boundary offset shift
```
johagge:
```suggestion
        :param offset: A vertical field boundary offset shift
```

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
JanGut: Practically, we could not shift the matrix, since, in my testing, it is not faster.

bitbots_vision/src/bitbots_vision/vision_modules/field_boundary.py
JanGut: Same here

Issue #177, opened 2019-12-12
Refactor imageloader


JanGut:
I suggest, refactoring the imageloader and making it a little more dynamic with more parameters (perhaps dyn reconf).

## Is your feature request related to a problem? Please describe.
- The imageloader is currently written in py2, must be moved to py3.
- Framerate, etc. is hardcoded.
- Source folder is not dynamic reconfigurable.

## Describe the solution you'd like
See above...

## Describe alternatives you've considered
None

PR #203, opened 2020-02-21
Better imageloader

Flova:
This adds a better image loader script to publish an image set as ROS messages.

## Proposed changes
- Add script
- Remove old imageloader package

## Related issues
- #177 

## Necessary checks
- [ ] Update package version
- [x] Run `catkin build`
- [x] Write documentation
- [ ] ~Create issues for future work~
- [x] Test on your machine
- [ ] ~Test on the robot~


bitbots_vision/scripts/bitbots_imageloader.py
```diff
@@ -0,0 +1,109 @@
...
+    import progressbar
+    progressbar_installed = True
+except ImportError:
+    progressbar_installed = False
+from sensor_msgs.msg import Image
+from cv_bridge import CvBridge, CvBridgeError
+
+
+class LoadImages:
+    def __init__(self, path, frame="/camera_link", topic="image_raw", fps=10, loop=False, pgbar=False):
```
NFiedler: I think that camera_optical_frame is better here.

bitbots_vision/scripts/bitbots_imageloader.py
```diff
@@ -0,0 +1,109 @@
...
+            for im in add_progress_bar(sorted(listdir)):
+                rospy.logdebug(im, logger_name="imageloader")
+
+                # Build path of the current image
+                image_path = os.path.join(path, im)
+                # Load image
+                image = cv2.imread(image_path)
+
+                # Check if this is a valid image
+                if image is not None:
```
NFiedler: Did you check if this fucks up everything when there is a non-image file (i.e. with an exception or something)

bitbots_vision/scripts/bitbots_imageloader.py
```diff
@@ -0,0 +1,109 @@
...
+if __name__ == "__main__":
+    # Parse cli
+    parser = argparse.ArgumentParser("Publish a image set as ros messages.")
+
+    parser.add_argument("-p", "--path", help="Input directory for the images", dest="path", type=str)
+    parser.add_argument("-fps", "--frames-per-second", help="Playback speed.", default=10, dest="fps", type=int)
+    parser.add_argument("-t", "--topic", help="ROS topic where the images are published to.", default="image_raw",
+        dest="topic", type=str)
+    parser.add_argument("-f", "--frame", help="The tf frame where the image should be published.", default="/camera_link",
+        dest="frame", type=str)
```
NFiedler: I also think that camera_optical_frame is better here too.

bitbots_vision/scripts/bitbots_imageloader.py
Flova: Yes, I agree with you.

bitbots_vision/scripts/bitbots_imageloader.py
Flova: Yes I checked this case. In this case the `imread` function of OpenCV returns `None` and this gets checked in line 63.

bitbots_vision/scripts/bitbots_imageloader.py
```diff
@@ -0,0 +1,109 @@
...
+        # Create CV bridge
+        self.bridge = CvBridge()
+
+        # List images in dir
+        listdir = list(os.listdir(path))
+
+        # Set fps
+        rate = rospy.Rate(fps)
+
+        # Make generator to determin if the loop runs once or infinite times
```
johagge:
```suggestion
        # Make generator to determine if the loop runs once or infinite times
```

bitbots_vision/scripts/bitbots_imageloader.py
```diff
@@ -0,0 +1,109 @@
...
+                    rate.sleep()
+                if rospy.is_shutdown():
+                    break
+            if rospy.is_shutdown():
+                break
+
+
+if __name__ == "__main__":
+    # Parse cli
+    parser = argparse.ArgumentParser("Publish a image set as ros messages.")
```
johagge:
```suggestion
    parser = argparse.ArgumentParser("Publish an image set as ROS messages.")
```

bitbots_vision/scripts/bitbots_imageloader.py
```diff
@@ -0,0 +1,109 @@
...
+
+    parser.add_argument("-p", "--path", help="Input directory for the images", dest="path", type=str)
```
johagge:
```suggestion
    parser.add_argument("-p", "--path", help="Input relative or absolute directory for the images. If none specified, current directory will be assumed", dest="path", type=str)
```
Assuming relative/absolute work, this makes it more clear :)

bitbots_vision/scripts/bitbots_imageloader.py
```diff
@@ -0,0 +1,109 @@
...
+    parser.add_argument("-fps", "--frames-per-second", help="Playback speed.", default=10, dest="fps", type=int)
+    parser.add_argument("-t", "--topic", help="ROS topic where the images are published to.", default="image_raw",
+        dest="topic", type=str)
+    parser.add_argument("-f", "--frame", help="The tf frame where the image should be published.", default="/camera_link",
+        dest="frame", type=str)
+    parser.add_argument("-nl", "--no-loop", help="Should the image sequence not be looped.", dest="loop", action='store_false')
```
johagge:
```suggestion
    parser.add_argument("-nl", "--no-loop", help="Set this flag, if the image sequence should not be looped.", dest="loop", action='store_false')
```

PR #205, opened 2020-02-28
Line LUTs

### LINKED ISSUE #182: Line color spaces
## Is your feature request related to a problem? Please describe.
Bad lines.

## Describe the solution you'd like
Line color lookup tables

Flova:
## Proposed changes
- Add LUTs as possible line color definition instead of HSV values.
- Add median filter on binary image to reduce noise

## Related issues
Closes #182 

## Necessary checks
- [x] Update package version
- [x] Run `catkin build`
- [ ] ~Write documentation~
- [ ] ~Create issues for future work~
- [x] Test on your machine
- [x] Test on the robot


bitbots_vision/src/bitbots_vision/vision_modules/ros_utils.py
```diff
@@ -180,10 +180,11 @@ def add_color_space_enum(cfg_type, package_path):
     # Get all files in this directory
     color_space_files = [file for file in color_spaces if os.path.isfile(os.path.join(color_spaces_path, file))]
     # Create list with a new enum item for each file
-    field_color_space_enum = [{'name': cs_file, 'value': cs_file, 'description': 'color space {}'.format(cs_file)} for cs_file in color_space_files]
+    color_space_enum = [{'name': cs_file, 'value': cs_file, 'description': 'color space {}'.format(cs_file)} for cs_file in color_space_files]
 
     # Add enums to configuration
-    _change_enum_items(cfg_type, 'field_color_detector_path', field_color_space_enum)
+    _change_enum_items(cfg_type, 'field_color_detector_path', color_space_enum)
+    _change_enum_items(cfg_type, 'white_color_detector_color_space_path', color_space_enum)
```
NFiedler: Does this actually work? It seems like you replace the same thing twice.

bitbots_vision/src/bitbots_vision/vision_modules/color.py
```diff
@@ -285,10 +288,10 @@ def update_config(self, config):
 
         super(PixelListColorDetector, self).update_config(config)
 
-        if ros_utils.config_param_change(tmp_config, config, 'field_color_detector_path'):
+        if ros_utils.config_param_change(tmp_config, config, self.color_space_path_param):
             # concatenate path to file containing the accepted colors of base color space
-            path = os.path.join(self._package_path, 'config/color_spaces')
-            color_space_path = os.path.join(path, config['field_color_detector_path'])
+            path = os.path.join(self._package_path, 'config', 'color_spaces')
+            color_space_path = os.path.join(path, config[self.color_space_path_param])
```
JanGut:
```suggestion
            color_space_path = os.path.join(path, config[self._color_space_path_param])
```

bitbots_vision/src/bitbots_vision/vision_modules/ros_utils.py
Flova: No this is correct. The command inserts the enum of possible color spaces into the drop-down menu for the field color spaces and the drop-down menu for the line colorspaces. This is the reason, why the same enum gets insertet into the drop down of two parameters.

bitbots_vision/src/bitbots_vision/vision_modules/lines.py
```diff
@@ -155,7 +155,9 @@ def get_line_mask(self):
             # Get not green points under field boundary
             possible_line_locations = cv2.bitwise_and(not_green_mask, not_green_mask, mask=field_boundary_mask)
             # Get white points that are not above the field boundary or in the green field
-            self._white_mask = self._white_detector.mask_bitwise(possible_line_locations)
+            white_mask = self._white_detector.mask_bitwise(possible_line_locations)
+
+            self._white_mask = cv2.medianBlur(white_mask, 3)
```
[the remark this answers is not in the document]
Flova: I told Jan and if I got it right also you that this parameter has been added. It is applied to the binary mask to filter outliers in the mask. A bigger value would destroy all lines and a smaller is not possible. Therefore I would keep this param not reconfigurable. And I am the last person who would deny the usage of dynamic reconfigure, but in this case, it is not necessary in my opinion.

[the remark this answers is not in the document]
NFiedler: Okay. But at least explain the motivation for this in a comment.

JanGut:
```suggestion
            # Filter out outliers
            self._white_mask = cv2.medianBlur(white_mask, 3)
```

PR #228, opened 2020-09-01
Gesture recognition

SammyRamone:
## Proposed changes
Gesture recognition from trainer. Currently WIP

## Necessary checks
- [ ] Update package version
- [ ] Run `catkin build`
- [ ] Write documentation
- [ ] Create issues for future work
- [ ] Test on your machine
- [ ] Test on the robot
- [ ] Put the PR on our Project board

PR #233, opened 2020-10-28
manually copy gesture package from other branch due to git issues

SammyRamone:
Gesture recognition from trainer. Currently WIP
Original commits on https://github.com/bit-bots/bitbots_vision/pull/228

- [ ] Update package version
- [ ] Run `catkin build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


SammyRamone:
I think this is currently not of interest and if we want to do something like this, we will probably do it differently now

PR #244, opened 2020-11-20
mask robot

SammyRamone:
## Proposed changes
This puts a mask over the parts of the robot that are in the image, so that they don't produce false positives.
Needs some more work and testing

## Necessary checks
- [ ] Update package version
- [ ] Run `catkin build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


Flova:
I would not merge this at the moment, because it is not needed currently. But nevertheless, it would be cool to keep it if it's needed for something in the future.

jaagut:
I believe, we can now safely close this PR and remove the code. What are your opinions?

PR #245, opened 2020-11-20
Add python look up 

SammyRamone:
## Proposed changes
Uses Python instead of C++. Minimal less performant but gets rid of the C++.
Needs more work.

## Necessary checks
- [ ] Update package version
- [ ] Run `catkin build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_vision/src/bitbots_vision/vision_modules/color.py
+++ b/bitbots_vision/src/bitbots_vision/vision_modules/color.py
@@ -3,7 +3,6 @@ import cv2
 import yaml
 import pickle
 import rospy
-import VisionExtensions
 import numpy as np
 import os
 from cv_bridge import CvBridge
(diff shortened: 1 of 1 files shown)


Flova:
This mainly needs a discussion with our vision team. What do you think @NFiedler @jaagut @hendrikbrandt ?

NFiedler:
I would vote for more C++ in the future instead of less.

Flova:
@jaagut, @hendrikbrandt, @SammyRamone what do you think?

jaagut:
I am voting for moving this to python. It is way shorter and more readable.
Also we could remove our cpp extensions that additional and weird complexity.

In this case, we need to discuss, what to do with the clustering for the custom fcnn...
We could also try to move this to Python (as I remember correctly, @Flova once said, this can be done with a little bit of np magic).
Or we could drop the fcnn support entirely. We have not used it for over 1.5 years and the setup is most certainly not up to date with today's tensorflow standards.

Flova:
Dropping the fcnn would also shorten the code by a big margin and the startup will be improved by a huge margin due to the drop of the tensorflow dependency.

Flova:
Moving the clustering seems a bit tricky but I think there are many libs that do this. But on the other hand we aren't using the fcnn so

jaagut:
FCNN support is now removed with #281

Flova:
We could remove the cpp if the lookup is done in python.

PR #247, opened 2020-11-20
fix task scheduling

SammyRamone:
## Proposed changes
Same as https://github.com/bit-bots/bitbots_lowlevel/pull/72

## Necessary checks
- [x] Update package version
- [x] Run `catkin build`
- [x] Write documentation
- [x] Create issues for future work
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_vision/launch/vision_startup.launch
+++ b/bitbots_vision/launch/vision_startup.launch
@@ -7,10 +7,11 @@
     <arg name="debug" default="false" doc="true: activates publishing of several debug images" />
     <arg name="use_game_settings" default="false" doc="true: loads additional game settings" />
 
-
+    <arg if="$(optenv IS_ROBOT false)" name="taskset" default="taskset -c 6,7"/>
+    <arg unless="$(optenv IS_ROBOT false)" name="taskset" default=""/>
 
     <!-- Start the vision-->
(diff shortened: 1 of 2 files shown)


bitbots_vision/launch/vision_startup.launch
```diff
@@ -7,10 +7,11 @@
     <arg name="debug" default="false" doc="true: activates publishing of several debug images" />
     <arg name="use_game_settings" default="false" doc="true: loads additional game settings" />
 
-
+    <arg if="$(optenv IS_ROBOT false)" name="taskset" default="taskset -c 6,7"/>
```
Flova, line 10: Does the env variable `IS_ROBOT` exist at the moment? Because it would be also possible to check `ROBOT_NAME` variable which also only exists if there is a robot.

timonegk: We added `IS_ROBOT` to the Ansible. We also tried to use `ROBOT_NAME`, but the `if` in launch files requires a bool, a string gives a type error and is not casted to true.

Flova: Ah okay

Flova (approved):

NFiedler (approved):
Really interesting! 😅

SammyRamone:
Merged it for now. We can tweak a bit which part gets how many cores during next integration test

PR #266, opened 2021-01-27
yolodocs: add a troubleshooting option

johagge:
## Proposed changes

During training we had this issue multiple times, so it makes sense to explain how to fix it.

## Necessary checks
- [ ] Update package version
- [ ] Run `catkin build`
- [X] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot


jaagut:
maybe also add docs regarding `CUDA_VISIBLE_DEVICES`?

johagge:
I added another line to explain how to use a specific GPU with the framework itself, instead of how to do it with with CUDA_VISIBLE_DEVICES, because I have only used it in this way until now. Additionally, this also explains how to use both GPUs to speed up training.

jaagut (approved):

PR #293, opened 2021-10-28
ROS 2

Flova:
## Proposed changes
Move to ROS 2, so there are too many changes, maybe I sum them up at a later stage.


bitbots_vision/bitbots_vision/vision.py
```diff
@@ -92,64 +89,32 @@ def __init__(self):
...
-                except rospy.exceptions.ROSTimeMovedBackwardsException:
-                    pass
-
-    def _dynamic_reconfigure_callback(self, config, level):
+    def _dynamic_reconfigure_callback(self, params):
         """
         Callback for the dynamic reconfigure configuration.
 
         :param config: New _config
         :param level: The level is a definable int in the Vision.cfg file. All changed params are or ed together by dynamic reconfigure.
```
johagge, line 109: the parameters for the function have changed, so these docs are outdated now

bitbots_vision/bitbots_vision/vision.py
```diff
@@ -410,19 +375,18 @@ def _image_callback(self, image_msg):
...
 
         if self._transfer_image_msg_mutex.locked():
             return
 
         with self._transfer_image_msg_mutex:
-            # Transfer the image to the main thread
-            self._transfer_image_msg = image_msg
+            self._handle_image(image_msg)
 
+    @profile
```
johagge, line 389: What happens if the import for profile failed, would this crash the vision?
(because profile is in a try/except block)

bitbots_vision/bitbots_vision/params.py
```diff
@@ -0,0 +1,178 @@
...
+
+from rcl_interfaces.msg import ParameterDescriptor, FloatingPointRange, IntegerRange, ParameterType
+
+class ParameterGenerator:  # TODO own file
+  def __init__(self):
+    self.param_cache = []
+
+  def declare_params(self, node):
+    for param in self.param_cache:
+      node.declare_parameter(*param)
```
timonegk, line 11:
```suggestion
    node.declare_parameter('', self.param_cache)
```

timonegk: See here: https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/node.py#L361 and the function above.

bitbots_vision/bitbots_vision/dynamic_color_lookup_table.py
```diff
@@ -2,7 +2,7 @@
 
 import cv2
 import yaml
-import rospy
+import rclpy
```
jaagut, line 5: Do we want to migrate this as well or are we ignoring this as we will probably move to YOEO?

Flova: I don't think we should put much effort into the dynamic colorspace.

bitbots_vision/setup.py
```diff
@@ -1,9 +1,39 @@
...
+package_name = 'bitbots_vision'
+
+
+setup(
+    name=package_name,
+    packages=find_packages(exclude=['test']),
+    data_files=[
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
+        ('share/' + package_name, ['package.xml']),
```
timonegk, line 16: I think it would be cleaner to use `os.path.join` here.

SammyRamone (approved):

Issue #294, opened 2021-11-24
Integrate YOEO in vision


timonegk:

Flova:
https://github.com/bit-bots/YOEO

SammyRamone:
will be closed by https://github.com/bit-bots/bitbots_vision/pull/297

PR #297, opened 2021-12-20
feature/yoeo_integration

phinik:
feature:
	- abstract YOEOHandlerTemplate, YOEODetector, YOEOSegmentationTemplate
	- YOEOHandlerPytorch implementation
	- YOEODetector and Segmentation implementations

## Proposed changes
added yoeo_handler.py which contains the pytorch yoeo handler so far

## Related issues
https://github.com/bit-bots/bitbots_vision/issues/294

## Necessary checks
- [ ] Test on your machine


bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+
+class YOEOHandlerPytorch(YOEOHandlerTemplate):
+    """
+    Using Pytorch to get YOEO predictions
+    """
+
+    def __init__(self, config, model_path):
+        super().__init__(config, model_path)
+
+        weights_path = join(model_path, "weights", "yoeo.pth")
```
Flova, line 118: See model import for the vision described above.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+        config_path = join(model_path, "config", "yoeo-rev-7.cfg")
+
+        self._model = torch_models.load_model(config_path, weights_path)
```
Flova, line 121: Nice namespacing, I like it!

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def _compute_segmentation_for(self, class_name: str) -> None:
+        segmentation = where(self._segmentation == self._seg_class_idx[class_name], 1, 0)
+        self._segmentations[class_name] = moveaxis(segmentation, 0, -1)
+
+    def set_image(self, img) -> None:
+        self._reset()
+        self._update_image(img)
+
+    def _reset(self) -> None:
```
Flova, line 97: This is pretty clean.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+    def get_candidates(self):
+        return self._yoeo_handler.get_detection_candidates_for("robot")
+
+
+class YOEOSegmentationTemplate(ABC):
+    def __init__(self, yoeo_handler: YOEOHandlerTemplate):
+        self._yoeo_handler = yoeo_handler
+
+    @abstractmethod
+    def get_segmentation(self):
```
Flova, line 182: In other parts of the vision `get_mask()` is used to describe this behavior. Especially in the field boundary detector. Maybe we need to fiddle a bit around with field boundary and line detectors too as we need to introduce common interfaces for lines and the field boundary mask. Currently, the field boundary is described by either a polygon line or a mask drawn by it and as YOEO is only producing masks it might result in issues.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+        ...
+
+
+class YOEOBackgroundSegmentation(YOEOSegmentationTemplate):
+    def __init__(self, yoeo_handler):
+        super().__init__(yoeo_handler)
+
+    def get_segmentation(self):
```
Flova, line 190: Same here and so on.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+        self._seg_class_idx = dict()
+
+        self._load_candidate_class_names(model_path)
+
+    def _load_candidate_class_names(self, model_path: str) -> None:
+        path = join(model_path, "data", "yoeo_names.yaml")
+        data = load_classes(path)
+
+        self._det_class_names = data['detection']
+        for key, i in zip(data['segmentation'], range(len(data['segmentation']))):
```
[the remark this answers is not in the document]
Flova: Instead of a dict one could also use the position in the list as it should be eqivalent to the mapping above.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+
+        self.predict()
+
+        if class_name not in self._segmentations.keys():
+            self._compute_segmentation_for(class_name)
+
+        return self._segmentations[class_name]
+
+    def _compute_segmentation_for(self, class_name: str) -> None:
+        segmentation = where(self._segmentation == self._seg_class_idx[class_name], 1, 0)
```
Flova, line 90: The query would look like this `self._seg_class_idx.index(class_name)` if a list instead of a dict is used.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+        self._det_confidence_thresh = config['yoeo_conf_threshold']
+
+        self._segmentation = None
+        self._segmentations = dict()
+        self._seg_class_idx = dict()
+
+        self._load_candidate_class_names(model_path)
+
+    def _load_candidate_class_names(self, model_path: str) -> None:
+        path = join(model_path, "data", "yoeo_names.yaml")
```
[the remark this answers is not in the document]
phinik: That's were the TORSO.names is (or better was) stored...

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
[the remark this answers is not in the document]
phinik: > One could use
> 
> ```python
> for i, key in enumerate(data['segmentation']):
> ```
> 
> and maybe also a dict comprehension:
> 
> ```python
> self._seg_class_idx = {key: i for i, key in enumerate(data['segmentation'])}
> ```
> 
> or is there something i am missing?

Good point, I'll change that.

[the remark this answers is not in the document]
phinik: > Instead of a dict one could also use the position in the list as it should be eqivalent to the mapping above.

Sure, but I think I list would just confuse. All we really need is the specified mapping. And at least theoretically, the complexity should be better with a dict (not that it would make that much of a difference with just 3 elements). I'd stick with a dict here if you don't mind?

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+logger = logging.get_logger('vision_yoeo')
+
+try:
+    from yoeo import models as torch_models, detect as torch_detect
+    from yoeo.utils.utils import load_classes
+except ImportError:
+    logger.error("Not able to import pytorchyolo. This might be fine if you use another method.")
+
+
+class YOEOHandlerTemplate(ABC):
```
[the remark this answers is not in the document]
phinik: I'm not a great fan of having a long inheritance pyramid tbh, especially not with abstract classes. Also, I couldn't even call `super().__init__(...)` for the `YoloHandler` as the it loads the class names "the yolo-way". Maybe it would be a better solution to define a common interface, from which both handlers inherit and which they implement. However, it's also a "not so nice" solution since YOEO would need a more advanced interface than YOLO (or two)...

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+
+        self._segmentation = None
+        self._segmentations = dict()
+        self._seg_class_idx = dict()
+
+        self._load_candidate_class_names(model_path)
+
+    def _load_candidate_class_names(self, model_path: str) -> None:
+        path = join(model_path, "data", "yoeo_names.yaml")
+        data = load_classes(path)
```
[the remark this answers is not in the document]
phinik: Fair point, I'll change that. Actually, that's how I inteded to do it in the beginning...

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
phinik: ?

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+class YOEOHandlerTemplate(ABC):
+    def __init__(self, config: dict, model_path: str):
+        self._use_caching = config['caching']
+
+        self._image = None
+        self._iou_non_max_suppression_thresh = config['yoeo_nms_threshold']
+
+        self._det_candidates = defaultdict(list)
+        self._det_class_names = None
+        self._det_confidence_thresh = config['yoeo_conf_threshold']
```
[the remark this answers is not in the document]
phinik: Yeah, I was unsure with how the config would look like for YOEO. But since there is a nms_threshold for YOEO, it probably makes sense to have a common prefix, don't you think?

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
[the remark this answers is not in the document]
Flova: Exactly, but this is only the case if the YOEO repository is present. If you install the yoeo python package only the code gets installed and the ROS node has no access to the `config`, `data` or `weights` folders. Instead the necessary files are shipped via data.bit-bots.de. A set of files for YOLO including the old `.names` file can be seen here: http://data.bit-bots.de/models/2021_04_22_jonas_torso21_full/

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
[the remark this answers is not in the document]
Flova: I agree that the complexity is most likely better with the dict, but for the object detector `self._det_class_names` is also only a list of names, which might be confusing, so I would like them to be the same data type. I know, that the mapping is accessed the other way around most of the time for the object detector, so it makes sense complexity wise, but it seems confusing nether the less.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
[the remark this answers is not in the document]
Flova: Yea I agree with you in that point.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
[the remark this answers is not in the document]
Flova: I agree with you.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
Flova: I wanted to say that the model seems to be loaded from the YOEO repo and not the models folder inside the vision which is downloaded from http://data.bit-bots.de/models/.

Flova: You can just do 

```python
weights_path = join(model_path, "yoeo.pth")
```

if I recall correctly.

phinik: Ahh, ok. I'll change that.

jaagut (commented):
I assume, the integration in the main vision pipeline and the necessary config changes are currently missing in this PR? Does the code already exist or do we need to produce it?

Besides, I really like your code! It is very precise and clean.

Flova:
> Do we want to store a names.yaml somewhere in our vision (/config or /models?) or do we want to integrate this config into our main config.yaml?

I think it should be shipped with the weights and model description as it is dependent on them.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,207 @@
...
+import cv2
+from numpy import where, moveaxis
+from os.path import join
+from rclpy import logging
+from typing import List
+
+logger = logging.get_logger('vision_yoeo')
+
+try:
+    from yoeo import models as torch_models, detect as torch_detect
```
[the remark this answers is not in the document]
Flova: I don't think so, as all are `yoeo_models` in this file and we might have different frameworks meaning `torch_models` is a better description.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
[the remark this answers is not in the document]
phinik: > Exactly, but this is only the case if the YOEO repository is present. If you install the yoeo python package only the code gets installed and the ROS node has no access to the `config`, `data` or `weights` folders. Instead the necessary files are shipped via data.bit-bots.de. A set of files for YOLO including the old `.names` file can be seen here: http://data.bit-bots.de/models/2021_04_22_jonas_torso21_full/ .

What path should I use then?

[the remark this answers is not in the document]
Flova: Just `join(model_path, "yoeo_names.yaml")` should be sufficient as the `model_path` points towards the location of the model in the `/models` folder in the vision.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
phinik: And what about the cfg file?

Flova: It is in the same folder and called `yoeo.cfg`

phinik:
> I assume, the integration in the main vision pipeline and the necessary config changes are currently missing in this PR? Does the code already exist or do we need to produce it?
> 
> Besides, I really like your code! It is very precise and clean.

Yeah, the PR only encompasses the handler (yet). What would have to be changed in the pipeline?

Flova:
First of all the handler needs to be initialized, but there are also some challenges as YOEO only outputs masks for the field boundary. Previously the field boundary was described by a polygon. One approach might be to use only masks for the field and refactor the other vision components to use it too.

jaagut:
Is this now ready for a full review?
I'd assume, in-depth testing will be done in #293?

phinik:
> Is this now ready for a full review? I'd assume, in-depth testing will be done in #293?

Most of it. The YOEOVision class is still work in progress as the other yoeo handlers (except for pytorch) do not exist yet.

jaagut:
What is the current status? Is this ready for testing and review?

phinik:
> What is the current status? Is this ready for testing and review?

Almost. Apart from some refactoring, there's onyl one little bug left to fix.

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,890 @@
+#! /usr/bin/env python3
```
jaagut, line 1: Do you include this shebang intentionally? I believe, this is not necessary as we do not want to execute this file standalone.

phinik: Nope, didn't add it (and haven't noticed it so far). Maybe PyCharm added it...

bitbots_vision/bitbots_vision/yoeo_vision.py
```diff
@@ -0,0 +1,257 @@
...
+
+        self._sub_image = None
+
+        # Reconfigure data transfer variable
+        self._transfer_reconfigure_data = None
+        self._transfer_reconfigure_data_mutex = Lock()
+
+        # Image transfer variable
+        self._transfer_image_msg = None
+        self._transfer_image_msg_mutex = Lock()
```
jaagut, line 57: This is probably/hopefully not necessary anymore, as in ROS2 the vision apparently now runs in a single threaded event-loop.

bitbots_vision/bitbots_vision/yoeo_vision.py
```diff
@@ -0,0 +1,257 @@
...
+            self._vision_components.append(YOEOLineDetectionComponent(self))
+        if new_config["component_field_boundary_detection_active"]:
+            self._vision_components.append(YOEOFieldBoundaryDetectionComponent(self))
+        if new_config["component_field_detection_active"]:
+            self._vision_components.append(YOEOFieldDetectionComponent(self))
+        if new_config["component_debug_image_active"]:
+            self._vision_components.append(DebugImageComponent(self))
+
+        for vision_component in self._vision_components:
+            vision_component.configure(new_config, self._yoeo_handler)
```
jaagut, line 179: Reminder: Why does every vision_component need a yoeo_handler?

bitbots_vision/bitbots_vision/yoeo_vision.py
```diff
@@ -0,0 +1,257 @@
...
+        self._yoeo_handler.set_image(image)
+        self._publish_image_to_components(image)
+
+        self._yoeo_handler.predict()
+        self._run_components(image_msg)
+
+    def _convert_image_msg_to_cv2_image(self, image_msg):
+        return self._cv_bridge.imgmsg_to_cv2(image_msg, 'bgr8')
+
+    def _publish_image_to_components(self, image) -> None:
```
jaagut, line 240: `Publish` is in the context of a pub/sub framework a bit misleading.
Instead, maybe:

- `_foreward_image_to_components`
- `_set_image_for_components`
- `_update_components_with_image`

bitbots_vision/bitbots_vision/yoeo_params.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
+
```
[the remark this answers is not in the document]
phinik: which vision params?

bitbots_vision/config/yoeo_visionparams.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+    obstacle_finder_method: 'convex'  # Method for the obstacle finder (distance, convex or step)
+    obstacle_color_threshold: 10  # An obstacle is defined as blue/red if it contains more blue or red than this threshold
+    obstacle_field_boundary_diff_threshold: 20  # Minimal distance between detected and convex field boundary to accept it as obstacle
+    obstacle_candidate_field_boundary_offset: 30  # Fixed height of obstacles above the field boundary
+    obstacle_candidate_min_width: 15  # Minimum width of an obstacle
+    obstacle_candidate_max_width: 150  # Maximum width of an obstacle
+    obstacle_finder_step_length: 5  # Length of an object detection step along the field boundary
+    obstacle_finder_value_increase: 2.0  # Factor of the impact of the height of the field boundary on the distance threshold
+    obstacle_own_team_color: blue  # current team color of our team (red or blue)
```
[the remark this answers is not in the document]
phinik: Looks like we forgot to remove it when we changed it at GORE

bitbots_vision/bitbots_vision/yoeo_vision.py
phinik: @Flova what's your opinion on this?

bitbots_vision/bitbots_vision/yoeo_vision.py
```diff
@@ -0,0 +1,257 @@
...
+        # Reconfigure data transfer variable
+        self._transfer_reconfigure_data = None
+        self._transfer_reconfigure_data_mutex = Lock()
+
+        # Image transfer variable
+        self._transfer_image_msg = None
+        self._transfer_image_msg_mutex = Lock()
+
+        self._yoeo_handler: Union[None, yoeo_handler.IYOEOHandler] = None
+        self._vision_components: Union[None, List[IVisionComponent]] = None
```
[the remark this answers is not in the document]
phinik: :D

bitbots_vision/bitbots_vision/yoeo_vision.py
phinik: Not every component does, but most do. Those which do not need it do not use it in their configure methods. There just has to be a common interface...

bitbots_vision/bitbots_vision/yoeo_vision.py
```diff
@@ -0,0 +1,257 @@
...
+    def _image_callback(self, image_msg) -> None:
+        # TODO: type
+        """
+        This method is called by the Image-message subscriber.
+        Too old Image-Messages are dropped.
+
+        Sometimes the queue gets too large, even if the size is limited to 1.
+        That is why we drop old images manually.
+        """
+        logger.debug(f"{self.__class__.__name__}._image_callback(...) called")
```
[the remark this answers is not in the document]
phinik: Seems like a leftover

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+        if self._camera_cap_could_be_on():
+            self._log_error()
+            self._issue_oral_warning()
+
+    def _camera_cap_could_be_on(self) -> bool:
+        mean_image_brightness = self._calculate_mean_image_brightness()
+        return mean_image_brightness < self._camera_cap_brightness_threshold
+
+    def _calculate_mean_image_brightness(self) -> float:
+        return sum(cv2.mean(self._image))
```
Flova, line 225: could be a single `self._image.mean()` it is not equivalent, but sufficient.

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+        )
+
+    def _filter_by_candidate_threshold(self, candidates: List[candidate.Candidate]) \
+            -> List[candidate.Candidate]:
+        return candidate.Candidate.rating_threshold(candidates, self._config['ball_candidate_rating_threshold'])
+
+    @staticmethod
+    def _create_ball_messages(candidates: List[candidate.Candidate]):
+        """
+        :rtype: List[soccer_vision_2d_msgs.msg._ball.Ball]  # TODO
```
Flova, line 310: Move this to a return typehint

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+        return list(map(ros_utils.build_ball_msg, candidates))
+
+    @staticmethod
+    def _create_balls_message(image_msg, ball_messages):
+        """
+        :param image_msg: Image message
+        :type image_msg:  sensor_msgs.msg._image.Image
+        :param ball_messages: List of ball messages
+        :type ball_messages: List[soccer_vision_2d_msgs.msg._ball.Ball]  # TODO
+        :rtype: soccer_vision_2d_msgs.msg._ball_array.BallArray  # TODO
```
Flova, line 321: TODO

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+        """
+        return ros_utils.build_ball_array_msg(image_msg.header, ball_messages)
```
Flova, line 323: This is a pretty simple one liner, I personally don't think these should be wrapped in another function if their behavior is obvious.

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+
+    def _publish_balls_message(self, balls_message) -> None:
+        """
+        :param balls_message: Balls message
+        :type balls_message: soccer_vision_2d_msgs.msg._ball_array.BallArray  # TODO
+        """
+        if balls_message:
```
Flova, line 330: Whats up with this if?

Flova: Is there a situation where there is no ball array message?

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+        :rtype: soccer_vision_2d_msgs.msg._goalpost_array.GoalpostArray  # TODO
+        """
+        return ros_utils.build_goal_post_array_msg(image_msg.header, goalpost_messages)
+
+    def _publish_goalposts_message(self, goalposts_message) -> None:
+        """
+        :param goalposts_message: Goalposts message
+        :type goalposts_message: soccer_vision_2d_msgs.msg._goalpost_array.GoalpostArray  # TODO
+        """
+        if goalposts_message:
```
Flova, line 429: Same here.

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+        :param goalpost_messages: List of goalpost messages
+        :type goalpost_messages: List[soccer_vision_2d_msgs.msg._goalpost.Goalpost]  # TODO
+        :rtype: soccer_vision_2d_msgs.msg._goalpost_array.GoalpostArray  # TODO
+        """
+        return ros_utils.build_goal_post_array_msg(image_msg.header, goalpost_messages)
+
+    def _publish_goalposts_message(self, goalposts_message) -> None:
+        """
+        :param goalposts_message: Goalposts message
+        :type goalposts_message: soccer_vision_2d_msgs.msg._goalpost_array.GoalpostArray  # TODO
```
Flova, line 427: There are many TODOs in the doc strings. I don't comment on each of them.

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+    @staticmethod
+    def _create_field_boundary_msg(image_msg, field_boundary_points):
+        """
+        :param image_msg: Image message
+        :type image_msg:  sensor_msgs.msg._image.Image
+        :param field_boundary_points: Points of field boundary
+        :type field_boundary_points: List[Tuple[int, int]]
+        :rtype: soccer_vision_2d_msgs.msg._field_boundary.FieldBoundary  # TODO
+        """
+        return ros_utils.build_field_boundary_msg(image_msg.header, field_boundary_points)
```
Flova, line 499: Same here, is this wrapper function needed?

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+        :type line_mask: numpy ndarray (height, width, channels)
+        """
+        self._debug_image.draw_mask(line_mask, color=DebugImageColors.lines)
+
+    def set_image(self, image) -> None:
+        """
+        :param image: Image to run the component on
+        :type image:  numpy nd.array (height, width, channels)
+        """
+        pass  # Intentional
```
Flova, line 593: Why is the pass here intentional?

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+
+    def _publish_balls_message(self, balls_message) -> None:
+        """
+        :param balls_message: Balls message
+        :type balls_message: soccer_vision_2d_msgs.msg._ball_array.BallArray  # TODO
+        """
+        if balls_message:
+            self._publisher.publish(balls_message)
+
+    def _add_candidates_to_debug_image(self, candidates: List[candidate.Candidate]) -> None:
```
Flova, line 333: Now all balls candidates are added, but the different filtering steps are shown via their thickness. Do you have an example image how this new behavior looks?

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+
+    def _add_convex_field_boundary_to_debug_image(self, convex_field_boundary_points: List[Tuple[int, int]]) -> None:
+        self._debug_image.draw_field_boundary(convex_field_boundary_points, DebugImageColors.field_boundary_convex)
+
+    def set_image(self, image) -> None:
+        """
+        :param image: Image to run the component on
+        :type image:  numpy nd.array (height, width, channels)
+        """
+        self._field_boundary_detector.set_image(image)
```
Flova, line 520: Why do you only set the field_boundary_detector image and not the image for the debug_image drawing? Same for the other set_image calls that have a debug image in their scope.

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -0,0 +1,889 @@
...
+        :type field_mask_msg: sensor_msgs.msg._image.Image
+        """
+        self._publisher.publish(field_mask_msg)
+
+    def set_image(self, image) -> None:
+        """
+        :param image: Image to run the component on
+        :type image:  numpy nd.array (height, width, channels)
+        """
+        pass  # Intentional
```
Flova, line 666: Same here

bitbots_vision/bitbots_vision/yoeo_vision.py
Flova: I see this similar to Jan. To be sure a single threaded executer could be used explicitly. This is different to ROS1 where the default executor was multi threaded.

bitbots_vision/bitbots_vision/yoeo_vision.py
```diff
@@ -0,0 +1,257 @@
...
+        self._configure_vision(new_config)
+        self._config = new_config
+
+        return SetParametersResult(successful=True)
+
+    def _get_updated_config_with(self, params) -> Dict:
+        new_config = deepcopy(self._config)
+        for param in params:
+            new_config[param.name] = param.value
+        return new_config
```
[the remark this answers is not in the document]
Flova: I think we have no nested params currently. But we could add them in the future.

bitbots_vision/bitbots_vision/yoeo_vision.py
```diff
@@ -0,0 +1,254 @@
...
+            self._config,
+            config,
+            self._sub_image,
+            'ROS_img_msg_topic',
+            Image,
+            callback=self._image_callback
+        )
+
+    def _image_callback(self, image_msg) -> None:
+        # TODO: type
```
Flova, line 190: TODO

bitbots_vision/bitbots_vision/yoeo_vision.py
```diff
@@ -0,0 +1,254 @@
...
+
+        image = self._extract_image_from_message(image_msg)
+        if image is None:
+            logger.error("Vision pipeline - Image content is None")
+            return
+
+        self._yoeo_handler.set_image(image)
+        self._forward_image_to_components(image)
+
+        self._yoeo_handler.predict()
```
Flova, line 231: Isn't this implicitly done by the first component that requests e.g. the balls?

Flova (changes requested):
Overall it looks pretty good to me imo.. I have some changes/comments but most are on the minor side. I like the layout, but in some cases some functions are wrapped a bit extensively but this is just my coding style opinion. Also some functions/classes are missing type hints and doc strings, but most of them look pretty good. There are some TODOs left in the code, it would be nice, if they are resolved, removed or converted into an issue. Thank you for your work!

bitbots_vision/bitbots_vision/vision_modules/ros_utils.py
```diff
@@ -323,6 +331,7 @@ def config_param_change(old_config, new_config, params_expressions, check_genera
             return True
     return False
 
+
 def publish_vision_config(config, publisher):
```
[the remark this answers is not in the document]
Flova: Not really related to the PR

bitbots_vision/bitbots_vision/vision_modules/ros_utils.py
```diff
@@ -366,8 +378,29 @@ def get_team_from_robot_color(color):
...
         own_color = _game_state.team_color
+    else: 
+        return Robot().attributes.TEAM_UNKNOWN
 
     if color == own_color:  # Robot is in own team, if same color
         return Robot().attributes.TEAM_OWN
     else:  # Robot is not same color, therefore it is from the opponent's team
         return Robot().attributes.TEAM_OPPONENT
+
+def get_robot_color_for_team(team):
```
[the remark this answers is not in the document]
Flova: Nvm is found the code

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler_utils.py
```diff
@@ -0,0 +1,462 @@
...
+        else:  # best class only
+            conf = np.max(class_confidence_scores, axis=1)
+            j = np.argmax(class_confidence_scores, axis=1)
+            x = np.concatenate((box_coordinates, conf[:, None], j[:, None]), axis=1)[conf > self._conf_thresh]
+
+        return x
+
+    @staticmethod
+    def _is_multilabel_prediction(class_confidence_scores) -> bool:
+        return class_confidence_scores.shape[1] > 1  # makes absolutely no sense
```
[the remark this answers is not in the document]
phinik: What I mean by this comment is that multilabel prediction is triggered as soon as there is more than one class, whilst single class prediction is only active if there is just one class. In case of single class prediction, a max is performed over the classes which makes no sense if there is just one class anyway. This is how the pytorch implementation does it. I thought about having another parameter in the config, like "yoeo_single_class_pred: true / false"

bitbots_vision/bitbots_vision/vision_modules/ros_utils.py
```diff
@@ -366,8 +373,27 @@ def get_team_from_robot_color(color):
...
+    # Color is known and we have to first figure out the own color
+    # Get own team color
+    own_color = None
+    if _game_state is not None:
+        own_color = _game_state.team_color
+    else: 
+        return
```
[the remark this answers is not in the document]
jaagut: Yes.

bitbots_vision/bitbots_vision/vision_modules/color.py
```diff
@@ -10,10 +10,13 @@
...
 from cv_bridge import CvBridge
 from bitbots_vision.vision_modules import ros_utils
+from .field_boundary import IFieldDetector
+
 
 logger = logging.get_logger('bitbots_vision')
 
-class ColorDetector(object):
+
+class ColorDetector(IFieldDetector):
```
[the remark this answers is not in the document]
jaagut: That's right, this is not intuitive. Moreover, the `IFieldDetector` is really small, so there is not a large benefit.

bitbots_vision/bitbots_vision/vision.py
```diff
@@ -22,7 +22,7 @@
 logger = logging.get_logger('bitbots_vision')
 
 try:
-    from profilehooks import profile, timecall # Profilehooks profiles certain functions in you add the @profile or @timecall decorator.
+    from profilehooks import profile, timecall # Profilehooks profiles certain functions if you add the @profile or @timecall decorator.
 except ImportError:
```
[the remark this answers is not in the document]
jaagut: I tested it once and nothing died.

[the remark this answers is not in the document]
phinik: I tested it months ago and it still worked. It also worked at GORE when you think back.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,589 @@
...
+
+
+class YOEOHandlerTemplate(IYOEOHandler):
+    def __init__(self, config: Dict, model_directory: str):
+        logger.debug(f"Entering YOEOHandlerTemplate constructor")
+
+        self._det_candidates: Dict = defaultdict(list)
+        self._det_class_names: Union[None, List[str]] = None
+
+        self._image = None  # : Union[None, numpy.ndarray]
```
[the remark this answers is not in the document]
phinik: Not with the used numpy version afaik.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler_utils.py
```diff
@@ -0,0 +1,462 @@
...
+
+        output = self._postprocess_nms_output(detections, indices)
+
+        return output
+
+    def _preprocess_detections_for_nms(self, detections):
+        detections = self._filter_by_objectness_confidence(detections)
+        detections = self._calculate_class_confidence_scores(detections)
+        detections = self._pin_down_last_dimension_to_6(detections)
+        #detections = self._filter_by_class_confidence(detections)
```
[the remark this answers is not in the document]
phinik: No, it can be removed. It was some sort of experiment.

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
jaagut: I asked myself the same. Please add a reason.

jaagut:
I resolved many comments and added type annotations. I believe most remaining comments are minor and we can merge this soon!

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -86,7 +86,7 @@ def __init__(self, config: Dict, model_directory: str):
         self._det_candidates: Dict = defaultdict(list)
         self._det_class_names: Union[None, List[str]] = None
 
-        self._image = None  # : Union[None, numpy.ndarray]
+        self._image: Union[None, np.ndarray] = None
```
phinik, line 89: Did you test this?

bitbots_vision/bitbots_vision/yoeo_vision.py
```diff
@@ -176,21 +177,18 @@ def _register_subscribers(self, config: Dict) -> None:
             callback=self._image_callback
         )
 
-    def _image_callback(self, image_msg) -> None:
-        # TODO: type
+    def _image_callback(self, image_msg: Image) -> None:
```
phinik, line 180: Same here, does it still run? Most "missing" type hints aren't missing because I forgot them...

bitbots_vision/bitbots_vision/vision_modules/color.py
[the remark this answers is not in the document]
phinik: I have no idea tbh. It's removed

bitbots_vision/bitbots_vision/vision_modules/obstacle.py
```diff
@@ -317,6 +325,8 @@ def set_image(self, image):
         :param image: image the current image vision
         """
         self._obstacle_detector.set_image(image)
+        if self._color_detector is not None:
+            self._color_detector.set_image(image)
```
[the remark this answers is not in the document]
phinik: It is. _misc_obstacles_detector in YOEOObstacleDetectionComponent doesn't have a color detector

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,589 @@
...
+        ...
+
+    @abstractmethod
+    def get_available_segmentation_class_names(self) -> List[str]:
+        ...
+
+    @abstractmethod
+    def get_segmentation_mask_for(self, class_name: str):
+        """
+        :rtype: numpy.ndarray(shape=(height, width, 1))
```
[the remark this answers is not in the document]
phinik: Will have a look at it, but low priority imho

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
```diff
@@ -0,0 +1,589 @@
...
+        return detections, segmentation
+
+
+class YOEOPathGetter:
+    @classmethod
+    def _assemble_full_path(cls, model_directory: str, subdir: Optional[str], filename: str) -> str:
+        if subdir is None:
+            path = os.path.join(model_directory, filename)
+        else:
+            path = os.path.join(model_directory, subdir, filename)
```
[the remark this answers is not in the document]
phinik: It became too messy with all the models...

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler_utils.py
```diff
@@ -0,0 +1,462 @@
...
+        if self._too_many_boxes_remain_in(detections):
```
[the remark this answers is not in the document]
phinik: Removed

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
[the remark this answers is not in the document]
jaagut: Not tested on the robot directly, but we presumably use version 1.21.5

bitbots_vision/bitbots_vision/vision_modules/yoeo_vision_components.py
```diff
@@ -1,12 +1,14 @@
...
+    from soccer_vision_2d_msgs.msg import Ball
 
 import cv2
 import numpy as np
 import rclpy
 from rclpy.node import Node
 from sensor_msgs.msg import Image
 from humanoid_league_msgs.msg import Audio, GameState
-from soccer_vision_2d_msgs.msg import Ball, BallArray, FieldBoundary, Goalpost, GoalpostArray, RobotArray, Robot
+from soccer_vision_2d_msgs.msg import BallArray, FieldBoundary, Goalpost, GoalpostArray, RobotArray, Robot
```
jaagut, line 11: Probably same for `Goalpost`

bitbots_vision/bitbots_vision/yoeo_vision.py
jaagut: This one definitely worked on the robot.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
jaagut: As the vision worked and did not unexpectedly die on the robot, I guess, this also worked.

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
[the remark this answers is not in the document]
Flova: Totally low prio, no need to change this in the pr

bitbots_vision/bitbots_vision/vision_modules/yoeo_handler.py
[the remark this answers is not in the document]
Flova: Why would `np.ndarray` need type support from NumPy? You can just add the python type of the object. The difference is that newer NumPy versions feature the `ArrayLike` type, which is something different. Uniform Lists etc. are `ArrayLike` objects for NumPy this means that they are convertible using `np.asarray`. The `ndarray` type is stricter as it expects a NumPy array. `ArrayLike` should mostly be used in combination with a `np.asarray` cast if e.g. a user provides some data that should be plotted or something like this, where one might simply use a list for short inputs. In most cases in the vision or everywhere where we handle images this is not the case, and we strictly expect a NumPy array, so it is fine to be explicit.

bitbots_vision/bitbots_vision/yoeo_vision.py
jaagut: Yes, but I like that it's done explicitly.

jaagut (approved):

jaagut:
@Flova you need to review again...

Flova (approved):

PR #315, opened 2022-07-09
Use param blackboard for team color

jaagut:
## Proposed changes
Use parameter_blackboard team_color instead of gamestate message


bitbots_vision/bitbots_vision/params.py
```diff
@@ -88,6 +88,8 @@ def add(self, param_name, param_type=None, default=None, description=None, min=N
 gen.add("ROS_field_mask_image_msg_topic", str, description="ROS topic of the field mask debug image message")
 gen.add("ROS_gamestate_topic", str, description="ROS topic of the GameState message")
 
+gen.add("use_game_settings", bool, description="uses game settings such as the own team color")
```
Flova, line 91: Still adds the `use_game_settings` param.

jaagut: Good catch! I thought, I already removed this.

Flova (approved):

PR #320, opened 2022-11-12
Use duplicated config instead of overrides from sim config

Flova:
- Also change obstacle to robot topic
- Remove unused config


bitbots_vision/config/visionparams_sim.yaml
```diff
@@ -0,0 +1,97 @@
...
+    field_boundary_detector_search_method: "downsampling_reversed" # Method for finding the field boundary (iteration, reversed, downsampling_reversed, binary)
+    field_boundary_detector_vertical_steps: 60 # Number of steps on each scanline
+    field_boundary_detector_horizontal_steps: 30 # Number of scanlines
+    field_boundary_detector_roi_height: 4 # Region Of Interest height in which we are looking for green
+    field_boundary_detector_roi_width: 2 # Region Of Interest width in which we are looking for green
+    field_boundary_detector_roi_increase: 0.1 # Value that increases the region of interest if it is located lower in the image
+    field_boundary_detector_green_threshold: 157 # Threshold of green in the area covered by the kernel
+    field_boundary_detector_head_tilt_threshold: 20 # Threshold for the dynamic search method, that describes the head angle at which we are switching between the iteration and the reversed search method.
+
+    line_detector_field_boundary_offset: -15 # Threshold in which we are also searching for lines over the field boundary
```
jaagut, line 77: Do you know why we have chosen this value?
Is it to prevent detecting false positive lines, if the field boundary is slightly too high?

bitbots_vision/config/yoeo_visionparams_sim.yaml
```diff
@@ -0,0 +1,69 @@
+bitbots_vision:
+  ros__parameters:
+    # Defining the default vision parameters
```
jaagut, line 3:
```suggestion
    # Defining the simulation parameters for YOEO vision
```
Same for normal YOEO params

bitbots_vision/config/visionparams_sim.yaml
Flova: I think so

jaagut (approved):

PR #325, opened 2023-02-21
Vision: Fix update own team color call

### LINKED ISSUE #324: Legacy vision dies when detecting couloured obstacles

jaagut:
Might fix #324


Flova:
I think this fix is correct, but it does not fix the issues we had when the vision died after seeing an obstacle, as this issue here is in the initialization.

jaagut:
Obsolete, as we will be removing the non-YOEO vision shortly

PR #328, opened 2023-03-06
optimized HSV-Detectors

phinik:
## Proposed changes
- TORSO-21 optimized detector values
- 2nd red detector to take the zero-crossing from 180 to 0 into account

## Necessary checks
- [ ] Update package version


bitbots_vision/bitbots_vision/vision_modules/color.py
```diff
@@ -253,7 +313,13 @@ def _mask_image(self, image):
         :return np.array: masked image
         """
         hsv_image = cv2.cvtColor(image, cv2.COLOR_BGR2HSV)
-        return cv2.inRange(hsv_image, self._min_vals, self._max_vals)
+
+        if self._zero_crossing:
+            interval_A = cv2.inRange(hsv_image, self._min_vals_A, self._max_vals_A)
+            interval_B = cv2.inRange(hsv_image, self._min_vals_B, self._max_vals_B)
+            return np.clip(interval_A + interval_B, 0, 255)
```
Flova, line 320: We could just do 

```python
return interval_A | interval_B
```

It also works on arrays. It is a binary operator, but the values returned by `cv2.inRange` are `255`, so it doesn't really matter.

bitbots_vision/bitbots_vision/vision_modules/obstacle.py
```diff
@@ -354,7 +354,11 @@ def get_candidates(self):
                             obstacle.get_upper_left_x():obstacle.get_lower_right_x()
                         ]
                     )
-                    # Add obstacles over the threshold to the output list
+
+                    # Normalize colorness to per mille
+                    colorness = colorness * 1000 / 255
```
Flova, line 359: Why do we divide by 255?

bitbots_vision/bitbots_vision/vision_modules/color.py
phinik: Good idea, will change that.

bitbots_vision/bitbots_vision/vision_modules/obstacle.py
Flova: Ah nvm

Flova (approved):

PR #330, opened 2023-04-06
renamed parameter "obstacle_team_color_detection" 

phinik:
renamed parameter "obstacle_team_color_detection" into "yoeo_team_colors_provided" as it is rather a feature of the utilized YOEO than of the vision

### DIFF
--- a/bitbots_vision/bitbots_vision/yoeo_params.py
+++ b/bitbots_vision/bitbots_vision/yoeo_params.py
@@ -96,6 +96,7 @@ gen.add("yoeo_model_path", str, description="Name of YOEO model")
 gen.add("yoeo_nms_threshold", float, description="YOEO Non-maximum suppression threshold", min=0.0, max=1.0)
 gen.add("yoeo_conf_threshold", float, description="YOEO confidence threshold", min=0.0, max=1.0)
 gen.add("yoeo_framework", str, description="The neural network framework that should be used ['pytorch', 'openvino', 'onnx', 'tvm']")
+gen.add("yoeo_team_colors_provided", bool, description="Must be true if yoeo provides team color predictions, otherwise false")
 
 gen.add("ball_candidate_field_boundary_y_offset", int, description="Threshold in which ball candidates over the field boundary are allowed.", min=0, max=800)
 gen.add("ball_candidate_rating_threshold", float, description="A threshold for the minimum candidate rating", min=0.0, max=1.0)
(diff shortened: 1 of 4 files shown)


bitbots_vision/bitbots_vision/yoeo_params.py
```diff
@@ -96,6 +96,7 @@ def add(self, param_name, param_type=None, default=None, description=None, min=N
 gen.add("yoeo_nms_threshold", float, description="YOEO Non-maximum suppression threshold", min=0.0, max=1.0)
 gen.add("yoeo_conf_threshold", float, description="YOEO confidence threshold", min=0.0, max=1.0)
 gen.add("yoeo_framework", str, description="The neural network framework that should be used ['pytorch', 'openvino', 'onnx', 'tvm']")
+gen.add("yoeo_team_colors_provided", bool, description="Must be true if YOEO provides team color predictions, otherwise false")
```
jaagut, line 99: What does "must be" mean? Do we toggle this, if we _want_ to use the team colors, or do we get unwanted behavior or errors if this bool does not match the yoeo-model?

phinik: In brief: it's gonna crash if the parameter is not set according to the used YOEO model.

Team colors are based on a subdivision of the robot class into three separate classes. Hence, if YOEO team colors should be used, YOEO must provide these three classes. On the other hand, if YOEO does provide these classes and you wanted to use the legacy HSV detectors instead (there is hardly any reason why you would want to do that), it's not gonna work because the HSV detectors require a single robot class in order to fetch the candidates correctly. So if team colors are provided by YOEO, it's team color detection MUST be used. If you want to use the HSV detectors, a YOEO without team colors MUST be used. You cannot switch it on or off without exchanging the model. The parameter is basically a switch telling the vision to expect one (value false) or three (value true) robot classes. Therefore, the proposed change.

jaagut: Okay, got it! Thank you for your explanation.
I agree with your renaming, it makes the situation definitely more clear. But I am a bit hesitant, as having to set this parameter in accordance with the model sounds like a source for potential human error. I know, it not that big a deal, as it's easy to test for and potentially, we could warn the user about the wrong parameter.
I wonder if it is possible without too much hassle to eliminate this potential problem by automatically selecting the correct setting.

Do the class-names for YOEO with team-color detection contain the colors? If that is the case, we could make the vision behave correctly dependent on this feature/metric.

jaagut (approved):

bitbots_vision/bitbots_vision/yoeo_params.py
phinik: Yeah, they do. However, it's tricky to automatically choose the right setting based on the class names. The class names are loaded by the YOEO handlers, but the configuration information would be required earlier when setting up the vision components. One could make the YOEO handlers provide this information to the vision setup process, but to be honest, I am not a great fan of this idea. Likewise, loading the class names in the vision setup process isn't a great choice either.

What could make sense -- also for future variations of YOEO -- is to have each model come with a separate config file containing such information. That way one could configure the vision and also allow for using YOEOs without segmentation head, depth head, etc. (whatever is to come). Although the latter should be quite painful to implement because I reckon there would be a need for one set of handlers per configuration.

Flova (approved):

PR #331, opened 2023-04-06
Fix/nms across multiple robot classes

phinik:
If multiple robot classes are present, non maximum suppression has to be performed across all robot classes instead of class-wise.

### DIFF
--- a/bitbots_vision/bitbots_vision/vision_modules/yoeo/utils.py
+++ b/bitbots_vision/bitbots_vision/vision_modules/yoeo/utils.py
@@ -276,11 +276,18 @@ class DefaultDetectionPostProcessor(IDetectionPostProcessor):
                  image_preprocessor: IImagePreProcessor,
                  output_img_size: int,
                  conf_thresh: float,
-                 nms_thresh: float):
+                 nms_thresh: float,
+                 robot_class_id_low: int,
+                 robot_class_id_high: int):
         self._image_preprocessor: IImagePreProcessor = image_preprocessor
         self._output_img_size: int = output_img_size
         self._conf_thresh: float = conf_thresh
         self._nms_thresh: float = nms_thresh
+
+        # these values are needed in order to perform a proper nms if multiple robot classes exist, i. e. if nms shall
+        # be performed across all robot classes and not per robot class
+        self._robot_class_id_low = robot_class_id_low
+        self._robot_class_id_high = robot_class_id_high
 
         self._nms_max_number_of_boxes = 30000
         self._nms_max_number_of_detections_per_image = 30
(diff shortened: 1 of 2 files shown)


bitbots_vision/bitbots_vision/vision_modules/yoeo/utils.py
```diff
@@ -276,12 +276,17 @@ def __init__(self,
...
                  conf_thresh: float,
-                 nms_thresh: float):
+                 nms_thresh: float,
+                 first_robot_class_id: int):
         self._image_preprocessor: IImagePreProcessor = image_preprocessor
         self._output_img_size: int = output_img_size
         self._conf_thresh: float = conf_thresh
         self._nms_thresh: float = nms_thresh
 
+        # this value are needed in order to perform a proper nms if multiple robot classes exist, i. e. if nms shall
```
jaagut, line 286:
```suggestion
        # This value is needed in order to perform a proper nms if multiple robot classes exist, i. e. if nms shall
```

bitbots_vision/bitbots_vision/vision_modules/yoeo/utils.py
```diff
@@ -277,15 +280,15 @@ def __init__(self,
...
                  nms_thresh: float,
-                 first_robot_class_id: int):
+                 robot_class_ids: List[int]):
         self._image_preprocessor: IImagePreProcessor = image_preprocessor
         self._output_img_size: int = output_img_size
         self._conf_thresh: float = conf_thresh
         self._nms_thresh: float = nms_thresh
 
-        # this value are needed in order to perform a proper nms if multiple robot classes exist, i. e. if nms shall
+        # These values are needed in order to perform a proper nms if multiple robot classes exist, i. e. if nms shall
```
Flova, line 289: The line break is a bit unlucky

bitbots_vision/bitbots_vision/vision_modules/yoeo/utils.py
```diff
@@ -360,12 +370,24 @@ def _pin_down_last_dimension_to_6(self, detections: np.ndarray) -> np.ndarray:
         return x
 
     def _get_boxes_and_scores_for_nms(self, detections: np.ndarray) -> Tuple:
-        class_offsets = detections[:, 5:6] * self._nms_max_width_height_in_pixels
+        if self._robot_class_ids:
+            class_offsets = np.where(self._is_robot_class(detections), self._robot_class_id(), detections[:, 5:6])
```
Flova, line 374: The substitutes the ids of the second/third robot class with the one if the first class for the nms. But this only applies for the `class_offsets` and does not modify the class in the detection / nms result. Correct?

Flova: Also we might not want to prefer the unknown robot color over the red/blue/... one in the nms. But this might a bit hard to implement.

Flova (approved):

bitbots_vision/bitbots_vision/vision_modules/yoeo/utils.py
phinik: > The substitutes the ids of the second/third robot class with the one if the first class for the nms. But this only applies for the `class_offsets` and does not modify the class in the detection / nms result. Correct?

Yes. Only the offsets are manipulated. All robot classes are now shifted by the same amount, i. e. to the same location in the image plane.



> Also we might not want to prefer the unknown robot color over the red/blue/... one in the nms. But this might a bit hard to implement.

Why do you think this would happen? It's actually the red class that is used to calculate the offsets, but it should not play a role anyway since only the offsets are manipulated. Or am I missing sth?

Flova: No this was just a general thought not related directly to the offsets. But if the network outputs a box for an unknown robot as well as a slightly different red one there is a chance we pick the unknown one even if the robot is probably red. But I think this case is not very likely and quite difficult to catch.

phinik: Ah now I understand. But shouldn't that only happen if the confidence of the unknown robot is higher? Hence, the error would result from a misclassification by YOEO and not from the nms.

Flova: That is correct. Be we could recover "some" information with it, but I do not think it really matters.

PR #334, opened 2023-04-29
Feature/yoeo config file

phinik:
## Proposed changes
Turn yoeo_names.yaml into an actual config file named model_config.yaml. model_config.yaml does not only provide the class names but also allows for setting whether a specific YOEO variant supports team color detection.

In the future, further parameters could be added to model_config.yaml if required.

## Necessary checks
- [ ] Test on your machine
- [ ] Test on the robot


bitbots_vision/bitbots_vision/vision_modules/yoeo/model_config.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
```
jaagut, line 1: Why did you import future?

bitbots_vision/bitbots_vision/vision_modules/yoeo/model_config.py
```diff
@@ -0,0 +1,32 @@
...
+            self._config = yaml.safe_load(f)
+
+    def get_detection_classes(self) -> List[str]:
+        return self._config['detection']['classes']
+
+    def get_segmentation_classes(self) -> List[str]:
+        return self._config['segmentation']['classes']
+
+    def team_color_detection_supported(self) -> bool:
+        return self._config['detection']['team_colors'] if 'team_colors' in self._config['detection'].keys() else False
```
jaagut, line 24: This is a bit unintuitive. Couldn't you solve it with:
```py
return 'team_colors' in self._config['detection'].keys() and self._config['detection']['team_colors']
```
or alternatively:
```py
return self._config['detection'].get('team_colors', False)
```

bitbots_vision/bitbots_vision/vision_modules/yoeo/yoeo_handlers.py
```diff
@@ -501,10 +490,6 @@ def _assemble_full_path(cls, model_directory: str, subdir: Optional[str], filena
 
         return path
 
-    @classmethod
-    def get_names_file_path(cls, model_directory) -> str:
-        return cls._assemble_full_path(model_directory, None, "yoeo_names.yaml")
-
```
jaagut, line 507: This method is not needed anymore?

phinik: That's correct. Loading the new model_config.yaml is solely handled by the ModelConfig class. Moreover, there is no "yoeo_names.yaml" anymore...

bitbots_vision/bitbots_vision/vision_modules/yoeo/model_config.py
```diff
@@ -0,0 +1,32 @@
+import yaml
```
Flova, line 1: This is still yaml

Flova: Nvm I saw that you switched back

Flova (approved):

PR #341, opened 2023-07-02
Only import the used YOEO runtimes

timonegk:


bitbots_vision/bitbots_vision/vision_modules/yoeo/yoeo_handlers.py
```diff
@@ -402,6 +386,7 @@ def model_files_exist(model_directory: str) -> bool:
                os.path.exists(YOEOPathGetter.get_pytorch_pth_file_path(model_directory))
 
     def _compute_new_prediction_for(self, image: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+        from yoeo import detect as torch_detect
```
Flova, line 389: Is there any overhead from importing torch etc. every frame?

timonegk: There is no big overhead, the module is not imported again. However, it is another python instruction that is executed, it takes about as much time as a print statement. We can still move it to the constructor.

bitbots_vision/bitbots_vision/vision_modules/yoeo/yoeo_handlers.py
```diff
@@ -383,6 +372,10 @@ def __init__(self,
         config_path = YOEOPathGetter.get_pytorch_cfg_file_path(model_directory)
         weights_path = YOEOPathGetter.get_pytorch_pth_file_path(model_directory)
 
+        from yoeo import models as torch_models
+        from yoeo import detect as torch_detect
```
jaagut, line 376: Please consider including the log message in the try/except block

Flova (approved):

jaagut (approved):

PR #343, opened 2023-07-04
Add image republishing for rosbags for wm 2023

jaagut:
## Proposed changes
Previously, we used the throttle node from topic_tools to reduce the frequency of logged images to 1 Hz. Unfortunately, adding this additional subscriber reduces the /camera/image_proc frequency by 30%. To avoid this additional subscription of the throttle node, we republish it in the vision with lower frequency.

## Related issues
https://github.com/bit-bots/bitbots_misc/pull/230


jaagut:
Tested and works. With this, we record an image approx. every 2.5 seconds

timonegk (approved):

Issue #348, opened 2023-09-18
Deactivate field boundary detection


phinik:
## Describe the solution you'd like
Deactivate the field boundary detection

jaagut:
Closed with #349

PR #349, opened 2023-09-18
remove field boundary detection from yoeo vision

phinik:
## Proposed changes
Remove the field boundary detection from yoeo vision

## Related issues
Solves https://github.com/bit-bots/bitbots_vision/issues/348

## Necessary checks
- [ ] Test on your machine
- [ ] Test on the robot

### DIFF
--- a/bitbots_vision/bitbots_vision/vision_modules/yoeo/__init__.py
+++ b/bitbots_vision/bitbots_vision/vision_modules/yoeo/__init__.py
@@ -1,4 +1,4 @@
 from .object_manager import YOEOObjectManager
 from .vision_components import IVisionComponent, YOEOComponent, CameraCapCheckComponent, BallDetectionComponent, \
-    GoalpostDetectionComponent, FieldBoundaryDetectionComponent, LineDetectionComponent, \
+    GoalpostDetectionComponent, LineDetectionComponent, \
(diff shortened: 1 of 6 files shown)


jaagut:
The original goal of the issue #348 was to remove the complete handling and detection of field boundaries, as they are not used and needed anymore. Thus, reducing code complexity in the vision node.
Another thing, but rather part of general cleanup is the removal of the not-YOEO vision.

phinik:
> The original goal of the issue #348 was to remove the complete handling and detection of field boundaries, as they are not used and needed anymore. Thus, reducing code complexity in the vision node. Another thing, but rather part of general cleanup is the removal of the not-YOEO vision.

I would prefer to have a dedicated issue + PR for removing the legacy vision. Removing the field boundaries alone would most likely break the legacy vision, and there is no point in untangling it from the legacy vision if the latter is to be removed anyway in a subsequent PR.

jaagut:
In favor of #351, this is obsolete.

phinik:
I actually meant we take this PR to remove the field boundaries from the YOEO vision and a second PR to remove the legacy vision in one go.

phinik:
Ready to review

jaagut (approved):
Tested it and seems to work!

PR #357, opened 2023-11-24
Cleanup/remove record bag

### LINKED ISSUE #332: Move `record_bag.launch` to bitbots_misc
We should move the `record_bag.launch` file to bitbots_misc/bitbots_basler_camera if we want to keep it at all

jaagut:
# Summary
In #332 initially, we wanted to move the launch file to the `bitbots_basler_camera` package.
Then I noticed, that the static output file in `/tmp/` is not ideal, therefore tried to copy and adapt the large `rosbag_record.launch.py`.
Finally, I decided that would be too much effort, as we can just use that launch script or the simple CLI `ros2 bag record`.

## Proposed changes
- Removes old `record_bag.launch` file.
- Also updates documentation link in README.

## Related issues
Resolves #332

## Checklist

- [ ] Run `colcon build`
- [ ] Write documentation
- [ ] Test on your machine
- [ ] Test on the robot
- [X] This PR is on our `Software` project board

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ This is the vision ROS package of the Hamburg Bit-Bots.
 
 The vision is able to detect balls, lines, the field itself, the field boundary, goal posts, teammates, enemies and other obstacles.
 
-For further information and getting started, see the documentation of this package at our website: [doku.bit-bots.de](http://doku.bit-bots.de/package/bitbots_vision/latest/index.html)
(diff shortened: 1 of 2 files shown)


Flova (approved):

Flova:
Are there any references left in the documentation?

Flova:
I had a quick look and I don't think so

Flova:
We should be good to go

